PR #148, opened 2015-01-29
This resolves #147 - Update Maven properties documentation on how to set...

wimdeblauwe:
... the Maven project version as the revision

* Unit test added
* Readme updated

PR #149, opened 2015-01-29
fixing maven installation in AppVeyor

abelsromero:
The url used to download has been removed and this causes the build to fail.
It is only required to configure it using the new -Version parameter.


abelsromero:
BTW, I noticed that the badge in the README is not working properly. I'll replace it with the pulblic badge url format to simplify its management after the merge #148

mojavelinux:
:+1:

PR #151, opened 2015-02-01
fix #146 allowing Java extensions in the pom.xml

### LINKED ISSUE #146: Allow to set java extensions
Currently there’s no way to integrate java extensions ([asciidoctorj#extension-api](https://github.com/asciidoctor/asciidoctorj#extension-api)) in the maven plugin. So, my proposal is to include a new tag called `<processors>` (because there’s already an extension tag) to allow setting one or more  extensions in the pom.

Currently, there're 3 ways to register a extension, the first 2 (passing as String a fully qualified class and passing a Class object of the extension) are easy. For the 3rd one (passing an instance of the extension object) I’m thinking to allow writing Groovy code in the pom.xml itself of being able to reference a groovy script. This way it would be also possible to include inline extensions with the DSL.

The only problem I see is that the Groovy option may be missused for other purpouses, but otoh this provides a method to customize the build in ways that we cannot provide (see #144).

What do you think?

cc @lordofthejars

abelsromero:
This PR contains the implementations to allow configuring Java's AsciidoctorJ extensions in the pom according to the discussion on issue #146.
Also includes tests and the description of the new parameter `processors` in the README.adoc.

There're two thing I'd like to comment:
* A test to validate the configuration with the SPI interface (org.asciidoctor.extension.spi.ExtensionRegistry) is included. However, to avoid interference with other tests, it is disabled and the configuration file is located in a "wrong" location. There's a comment on the test describing how to activate it.
* In order to allow the configuration of BlockProcessor, BlockMacroProcessor or InlineMacroProcessor, which require an extra parameter, a new level on indentation in the XML has been added (see example in README). It's an extra layer of verbosity which I'm not proud off, but I did not see other options.

As always, comments are welcome.


mojavelinux:
Amazing work, @abelsromero!

> In order to allow the configuration of BlockProcessor, BlockMacroProcessor or InlineMacroProcessor, which require an extra parameter, a new level on indentation in the XML has been added (see example in README)

I'd actually say the nesting looks just as it should. It leaves room for present and future configuration for extensions, which are a critical part of making them useful.

I'm not sure if it's possible, but you could detect the presence of the nested elements and, if not there, assume the content is the class name. However, I'm happy with it if that's not feasible.

One question. How would you feel about the element `extensions` instead of `processors`. Although all the extensions are, in fact, processors, we've been referring to them simply as "extensions". wdyt?

abelsromero:
> I'm not sure if it's possible, but you could detect the presence of the nested elements and, if not there, assume the content is the class name. However, I'm happy with it if that's not feasible.

If you mean something like skipping the `processor` tag and only writting <className>...I already tried but it didn't work. I tested setting a second simple List<String> attribute, but having two parameters for the same thing was just weird and confusing.

> One question. How would you feel about the element extensions instead of processors. Although all the extensions are, in fact, processors, we've been referring to them simply as "extensions". wdyt?

This thing hit me all the time becase I had to keep remind me of the name `processor` and I'd be really happy to change it. But there is already an extensions attribute used to define file extensions (.adoc, .asciidoc, etc.), so I choose `processor` to keep it consistent with AsciidoctorJ API which calls them *Processors.
The thin is then, if we change that, what do we do with the other attribute? My approach would be in two steps:
1. (Now) Rename current extesions to fileExtensions. That's something we should document on some release notes to ease the change to the new version. Ppl upgrading and using the current 'extensions' will get a nasty error if they don't change the name in the pom.
2. (Future) Remove the current `extensions` and add a `sources` parameter based on patterns like the gradle plugin does (other maven plugins do that, so should be hard). This removes having similar attributes and adds more options.

Just let me know what's your opinion and I'd start.

abelsromero:
What do you think of the options? should I rename the other extensions?

mojavelinux:
(2) is the ideal solution. But I'm not sure that solves are name conflict. I would say that we should rename extensions to fileExtensions, but that does introduce the potential for some pain. However, I doubt that many people are using this option (I can't find anyone who is in all my checkouts), so perhaps it's enough just to mention the change in the release notes. Thus, I'm voting on the rename of `<extensions>` to `<fileExtensions>` and using `<extensions>` for the purpose defined in this pull request.

abelsromero:
Option 1 it is then :+1:
I'll get to it as soon as I can, which I am afraid won't be soon, srry :disappointed:

mojavelinux:
Get to it when you can. We're not going to disappear :)

abelsromero:
This is it, the final version. I had some problem with the merge and I could not squash everything, but I checked the changes done in the README and everything is in place.

mojavelinux:
@abelsromero This looks fantastic!

Before we merge, I'd like to reconsider the name for the option to configure the file extensions one more time. I checked the Gradle plugin and it uses the name `sourceDocumentPattern` to control the file extensions. Now, one main difference is that it allows an Ant-based pattern, not just a file extension name. But perhaps we can take a hint and call our option `sourceDocumentExtensions`. wdyt? Are we over thinking this, or do you see reason to align?

One reason to choose this name is that it makes it very explicit which files we are talking about. It also fits nicely next to `sourceDocumentName`. So maybe we are headed in the right direction with `sourceDocumentExtensions`.

abelsromero:
To some degree I think it's overthinking, but otoh I totally agree and defend aligning gradle and maven parameters :+1: I'll change `fileExtensions` for `sourceDocumentExtensions` and also move the explanation  in the readme right below `sourceDocumentName`.

Btw, I added this comment to the parameter `(named extensions in v1.5.3 and before)`, should it say before or below?

mojavelinux:
Cool!

I think where you mention the parameter change is fine. What's important is
that the information is there :)

mojavelinux:
:+1:

Issue #125, opened 2014-09-15
Maven site integration does not work with Maven Site Plugin 3.4


mosabua:
To reproduce just upgrade the version of the Maven Site Plugin in https://github.com/asciidoctor/asciidoctor-maven-examples/blob/master/asciidoc-site-example/pom.xml
and you will no longer get the html versions of your asciidoc generated..

mojavelinux:
I don't understand how to register multiple file extensions for a site module. If someone knows how to do it, please feel free to file a pull request. It will be most appreciated :)

paranoiabla:
You need to wait for the next release of the doxia plugin. Once that is done you will be able to simply do:
````
    /**
     * The extension for AsciiDoc files.
     */
    public static final String[] FILE_EXTENSION = { "adoc", "asciidoc" };

    /**
     * Build a new instance of {@link AsciidoctorParserModule}.
     */
    public AsciidoctorParserModule()
    {
        super( SOURCE_DIRECTORY, FILE_EXTENSION, AsciidoctorParser.ROLE_HINT );
    }
````

mojavelinux:
Great news!!

mosabua:
@paranoiabla what version of doxia? I might be able to push for a release on the Maven dev list. If its not already in progress..

paranoiabla:
The 2 pullrequests I have linked above (for `maven-doxia` and `maven-doxiasitetools`) need to be merged, a new release of doxia need to be issued. I believe @hboutemy is taking care of that :)

mstanleyjones:
I'm not sure if my problem is relating to this one. Using Maven site plugin 3.4, I can get .adoc files in src/site/asciidoc to build into standalone themed HTML files. However, if a file includes other files, the included files do not get resolved. For instance, in src/site/asciidoc/ I have book.adoc, which includes chapters/installation.adoc, chapters/administration.adoc, etc. If I build book.adoc using an execution of the asciidoctor plugin, it works. However, if I try to build it using mvn site, the includes do not get resolved. Presumably there is a step where everything gets concatenated into one big .adoc before building, and that is not happening with the 'mvn site' target.

I am happy to file this as a separate issue, but it seems that it might be related here so I thought I would post here first.

mojavelinux:
[...]

PR #152, opened 2015-02-14
Adding a new parser module seems to fix the multiple extensions.

paranoiabla:
#125 I'm not sure if this is allowed, or if it's a hack, but it seems to work for me. I have added another parser module for the `asciidoc` extension and now `mvn site:site` works for me.


paranoiabla:
Sorry about this - seems like adding a new parser will actually honor the `asciidoc` extension but ignore the `adoc` extension

Issue #156, opened 2015-04-15
Provide an option to suffix generated file names with pom version


matdere:
Hi,

First of all, thank you so much for making me discover asciidoctor during Devoxx France last week. You guys changed the way I will manage documentation now ! 

To my point of view, it would be great to have the possibilité to add the POM version to the generated files. 
The idea is to give the possibilité to the user to choose automatically the documentation that correspond to the software version they have. 

For example the file name would be : **mysupersoft-0.9.7.RELEASE.html**

What do you think about it ?

abelsromero:
Welcome Mathieu :smiley: 

This is not a bad idea, specially if you have several code branch (e.g. preview, stable, etc.) and you want to publish several versions of the documentation.
But I am not sure this fully solves the problem, specially when dealing with complex structures of documents with folders, images, css, etc. Also, browsing can be a bit complicated if the version is at the end of a folder structure in the file name. 
On the other hand, what do you think of setting the name in a folder? Currently, you can set the output folder in the `<configuration>` block in the following way:

```
<outputDirectory>target/generated-docs/${project.version}</outputDirectory>
```

What do you think?


PS: don't forget to check this https://github.com/asciidoctor/asciidoctor-maven-plugin#passing-pom-properties, to set the version in the document footer.

matdere:
Hola Abel, 
Awesome !!!! This is even better ! 
Thanks !

mojavelinux:
> thank you so much for making me discover asciidoctor during Devoxx France last week. You guys changed the way I will manage documentation now !

Excellent! I'm so glad to hear that!

> Currently, you can set the output folder in the <configuration> block in the following way...

Great suggestion Abel! Perhaps this is something we could document in the README / user manual because I bet a lot of people would like to use it. Perhaps we can add a "Tips & Tricks" section.

PR #157, opened 2015-04-28
Update README.adoc

matdere:
Following @mojavelinux  suggestion on the bug #156 : my little contribution to the project. Thanks again.


abelsromero:
Nice addition :) But I have a doubt, shouldn't `numbered` be in an `attributes` block? I can't test it right now. If that's the case, adding some extra information on how boolean properties are handled would be helpfull. There's already a small note in the README, but some more detail on the equivance of <numbered> with </numbered> would avoid confusions.
wdyt?

matdere:
@abelsromero you're right numbered should be an attibutes block. I did the modification and added an example to the boolean handling part also.

mojavelinux:
@matdere @abelsromero Even better would be to use the new name in 1.5.x for `numbered`, which is `sectnums`. This makes it more clear what is being numbered since it only applies to section numbers. If you can update, then otherwise the change looks great to me. Thanks for contributing it!

abelsromero:
Just out of curiosity, because I know you (Dan) are a very thoughtful person and I see you already corrected the user manual (https://github.com/asciidoctor/asciidoctor.org/commit/5457ae2f4e6ebbf65e7a66dd4ef4e47adba1cd43)

Is the attribute here mantained for some reason?
https://raw.githubusercontent.com/asciidoctor/asciidoctor.org/master/docs/_includes/ex-appendix.adoc

mojavelinux:
Thanks for pointing that out, @abelsromero. I probably just missed it. I was mostly focused on updating the main mention of it and I caught it a few other places. We need to keep going to weed all mentions out except the admonition note that says we still support `numbered` for backward compatibility.

PR #158, opened 2015-05-03
Remove the super call to the xhtml parser, because that one will call do...

paranoiabla:
...xia for every node which will cut out most of the attributes of different tags and will result in corrupt html. Instead we call sink.rawText() which will pass the already generated html.

### DIFF
--- a/src/main/java/org/asciidoctor/maven/site/AsciidoctorParser.java
+++ b/src/main/java/org/asciidoctor/maven/site/AsciidoctorParser.java
@@ -11,20 +11,19 @@
  */
 package org.asciidoctor.maven.site;
 
-import java.io.IOException;
-import java.io.Reader;
-import java.io.StringReader;
-
 import org.apache.maven.doxia.module.xhtml.XhtmlParser;
 import org.apache.maven.doxia.parser.ParseException;
 import org.apache.maven.doxia.parser.Parser;
 import org.apache.maven.doxia.sink.Sink;
 import org.asciidoctor.Asciidoctor;
-import org.asciidoctor.OptionsBuilder;
 import org.asciidoctor.AttributesBuilder;
+import org.asciidoctor.OptionsBuilder;
 import org.asciidoctor.SafeMode;
 import org.codehaus.plexus.component.annotations.Component;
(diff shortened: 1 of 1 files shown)


paranoiabla:
Guys, as discussed here: https://github.com/asciidoctor/asciidoctor/issues/947 the maven-doxia has a sink api which will trim a lot of the attributes from the html that asciidoctor generates. Thus:
````
<i class="fa icon-note" title="Note"></i>
````

becomes `<i></i>` because doxia thinks it is `<i>` means italics. Doxia is invoked when you calll `super()` on the `XhtmlParser`. To prevent this I use the method `rawText()` as suggested by @hboutemy and it seems to work like a charm. I also had to set `false` on the `.headerFooter(false)` because then headers and footers were included in the `<body>` of the final html. Now they're not and instead i have a custom maven theme which includes the font-awesome and the coreray css and it works great.

mojavelinux:
Most excellent! I didn't realize you could bypass the doxia parser and just send back the raw text. That's perfect! Now the documents are going to look the way they are supposed to look. Thanks for submitting this fix!

@abelsromero, I'd say we're good to go on this one.

>  I also had to set false on the .headerFooter(false) because then headers and footers were included in the <body> of the final html.

Yes, good. We needed the header/footer to make doxia happy, but now that we're using raw text it's correct to send back the embeddable HTML (which is headerFooter(false)).

> i have a custom maven theme which includes the font-awesome and the coreray css and it works great.

Fantastic!

abelsromero:
Thanks @mojavelinux, I was hoping for someone to lend a hand in validating this since the site mojo is a bit out of my scope currently ^^''

mojavelinux:
This should improve our site example. https://github.com/asciidoctor/asciidoctor-maven-examples/tree/master/asciidoc-maven-site-example There we'll be able to see the impact (and test it).

hboutemy:
there is only one drawback of using rawText(): in fact, it's bypassing Doxia site rendering but rendering HTML directly without Doxia, then pdf rendering won't work.

This is a limitation that should be clearly stated in the documentation (reasonable limitation IMHO, given what is permits in asciidoc case)

Issue #171, opened 2015-09-22
Address already in use error in the AsciidoctorHttpMojoTest


NicolasGeraud:
If port 2000 is used on during maven test phase, the build fail.

This port is the default port declared in the [AsciidoctorRefreshMojo](https://github.com/asciidoctor/asciidoctor-maven-plugin/blob/master/src/main/java/org/asciidoctor/maven/AsciidoctorRefreshMojo.java#L45).

It could be better to find a random available port on the host, and configure the test to use it.

There is still a risk that the port became used between the moment we find it and the moment we use it, but it's a more robust test I think.

Please share any suggestion before I start hacking it.

LightGuard:
Sounds good to me.

On Tue, Sep 22, 2015 at 3:12 PM, Nicolas Géraud <notifications@github.com>
wrote:

> If port 2000 is used on during maven test phase, the build fail.
>
> This port is the default port declared in the AsciidoctorRefreshMojo
> <https://github.com/asciidoctor/asciidoctor-maven-plugin/blob/master/src/main/java/org/asciidoctor/maven/AsciidoctorRefreshMojo.java#L45>
> .
>
> It could be better to find a random available port on the host, and
> configure the test to use it.
>
> There still a risk that the port became used between the moment we find it
> and the moment we use it, but it's a more robust test I think.
>
> Please share any suggestion before I start hacking it.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/asciidoctor/asciidoctor-maven-plugin/issues/171>.
>



-- 
Jason Porter
http://en.gravatar.com/lightguardjp

PR #172, opened 2015-09-22
fix(test): use a random available port on http test

NicolasGeraud:
Instead of using the default mojo port (2000) during http test, find the
first available port on the host an use it during tests.

#171


LightGuard:
Thanks!

mojavelinux:
Thanks @NicolasGeraud!

Issue #180, opened 2015-11-28
baseDir should be set when Asciidoctor is invoked from the Maven Site plugin


mojavelinux:
The baseDir option should be set when Asciidoctor is invoked from the Maven Site plugin. Currently, the baseDir is defaulting to the root of the project. This means that top-level includes have to be specified using a relative path from the project root instead of relative to the main file.

The behavior that needs to be emulated is specified here: https://github.com/asciidoctor/asciidoctor-maven-plugin/blob/master/src/main/java/org/asciidoctor/maven/AsciidoctorMojo.java#L213-L222

There's currently no way to provide options to the Maven Site plugin, so it should use the fallback behavior.

mstanleyjones:
Also, if you are trying to use maven-site with Asciidoc, but also to build a PDF payload, the include:: syntax for making the HTML works breaks the PDF.

mojavelinux:
Technically, there is a way to work around this problem, but I'm only mentioning it for completeness, not because it's elegant.

You can simple define an attribute as the root of the includes in the document header per backend.

```
ifdef::backend-pdf[]
:includedir: src/site/asciidoc
endif::[]
ifndef::backend-pdf[]
:includedir: .
endif::[]

include::{includedir}/part.adoc[]
```

Again, we really don't want to have to resort to this, but in the meantime it might help to get you by.

mojavelinux:
Btw, we should have an attribute set in the Maven Site configuration that indicates that we're in the maven site environment. That would make conditionals like this more clear. Something like `env-maven-site`.

mojavelinux:
Once I figured out how to inject the MavenProject into the class, I was able to access the basedir of the project and from there the path to the AsciiDoc files in the site. The basedir will be set to the following path by default:

* ${project.basedir}/src/site/asciidoc

This value can be configured using a configuration property under the site configuration.

```xml
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-site-plugin</artifactId>
    <version>3.3</version> <!-- 3.4 also works -->
    <configuration>
        <asciidoc>
            <baseDir>${project.basedir}</baseDir>
        </asciidoc>
    </configuration>
    <dependencies>
        <dependency>
            <groupId>org.asciidoctor</groupId>
            <artifactId>asciidoctor-maven-plugin</artifactId>
            <version>1.5.3</version>
        </dependency>
    </dependencies>
</plugin>
```

Issue #179, opened 2015-11-28
Allow templates directory to be specified for Maven Site plugin


mojavelinux:
Allow the templates directory to be specified in the configuration of the Maven Site plugin.

ORIGINAL DESCRIPTION: Select a standard directory where Asciidoctor templates are stored and enable the templates option when calling the convert method if this directory exists and contains files. This avoids the need to have to define the templates directory explicitly and allows templates to be used with the Maven Site plugin.

mojavelinux:
I propose the following two locations:

* src/docs/asciidoc/_templates
* src/site/asciidoc/_templates

I'm open to other ideas.

LightGuard:
Is there a standard config directory for maven projects?

On Friday, November 27, 2015, Dan Allen <notifications@github.com> wrote:

> I propose the following two locations:
>
>    - src/docs/asciidoc/_templates
>    - src/site/asciidoc/_templates
>
> I'm open to other ideas.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/asciidoctor/asciidoctor-maven-plugin/issues/179#issuecomment-160232468>
> .
>


-- 
Sent from Gmail Mobile

AlexanderZobkov:
I would propose to store Asciidoc templates in the following directories:
* ${basedir}/src/main/asciidoc/templates
* ${basedir}/src/site/asciidoc/templates

IMHO the "underscore" symbol looks extraneously for maven world, typical maven projects and typical directories layouts used by developers.
Regarding "src/docs", this is not in alligment with default directory for "sourceDirectory" configuration option: 
* defaults to ${basedir}/src/main/asciidoc

LightGuard:
src/main/asciidoc makes more sense for templates or extensions.

On Saturday, November 28, 2015, Alexander <notifications@github.com> wrote:

> I would propose to store Asciidoc templates in the following directories:
>
>    - ${basedir}/src/main/asciidoc/templates
>    - ${basedir}/src/site/asciidoc/templates
>
> IMHO the "underscore" symbol looks extraneously for maven world, typical
> maven projects and typical directories layouts used by developers.
> Regarding "src/docs", this is not in alligment with default directory for
> "sourceDirectory" configuration option:
>
>    - defaults to ${basedir}/src/main/asciidoc
>
> —
> Reply to this email directly or view it on GitHub
[...]

Issue #168, opened 2015-09-04
Maven Site does not work with .adoc where dependency depends from gem like ditaa extension


alexandrenavarro:
Maven Site does not work with .adoc where dependency depends from gem like ditaa extension

I try to generate a maven site a .adoc with dita.

When I used asciidoctor-maven-plugin directly with .adoc, it works but not with the maven-site.

To reproduce the problem take this sample

https://github.com/asciidoctor/asciidoctor-maven-examples/tree/master/asciidoctor-diagram-example

add in your pom.xml

			<plugin>
				<groupId>org.apache.maven.plugins</groupId>
				<artifactId>maven-site-plugin</artifactId>
				<version>3.4</version>
				<dependencies>
					<dependency>
						<groupId>org.asciidoctor</groupId>
						<artifactId>asciidoctor-maven-plugin</artifactId>
						<version>${asciidoctor.maven.plugin.version}</version>
					</dependency>
				</dependencies>
			</plugin>

and copy src/docs/asciidoc into src/site/asciidoc .

You will see

    [INFO] Rendering 1 Doxia document: 1 asciidoc
    asciidoctor: WARNING: <stdin>: line 18: invalid style for literal block: ditaa
    asciidoctor: WARNING: <stdin>: line 38: invalid style for literal block: plantuml

I think the problem is the maven-site-plugin does not know the dependency of the gem but I can't add the dependency of the gems (maven can't download) or add an equivalent of 

    <gemPath>${project.build.directory}/gems-provided</gemPath>
    <requires>
    <require>asciidoctor-diagram</require>
    </requires>

Tell me is a way to configure the maven-site to succeed to generate site or if it is a bug.

abelsromero:
This is a know issue realted to this https://github.com/asciidoctor/asciidoctor-maven-plugin/issues/160.

The thing is that the site module is a completely different component running under doxia API and we haven't figured out a nice way to pass configurations parameters to the site plugin.

Let's awaken the beast again and continue the conversation on the other side.

robertpanzer:
There is actually a way to work around this problem when using AsciidoctorJ 1.5.3.2:
You have to build an own artifact, let's say t's called `foo.bar:asciidoctor-site-diagram-activator:1.0`, that contains a service implementation of the class `org.asciidoctor.extension.spi.ExtensionRegistry`:
```java
package foo.bar;

import org.asciidoctor.Asciidoctor;
[...]

Issue #160, opened 2015-05-06
Allow the source highlighter to be enabled in the site plugin


mojavelinux:
It should be possible to enable the source highlighter from the plugin configuration when using the site plugin.

This issue involves two parts:

1. An option for enabling the source highlighter (similar to the main plugin)
2. Ensure that the source highlighting works when enabled (any necessary styles are included)

As a workaround, you can enable the source highlighter by adding an explicit attribute in the document header.

mojavelinux:
@paranoiabla, you mentioned you got this working in light of recent changes to the plugin. Could you elaborate?

paranoiabla:
I'm using the `:source-highlighter: coderay` option in the header of each asciidoctor file. That seems to work. The drawback is that if you want to change it, you have to go through all the files and manually change the header.

mojavelinux:
Is it possible to pass configuration to a Maven site module? If so, then we could pass that on to AsciidoctorJ.

abelsromero:
> Is it possible to pass configuration to a Maven site module?

According to documentation you can add a site descriptor (site.xml in src/site) to customize the layout and other things ([ref](http://maven.apache.org/plugins/maven-site-plugin/examples/sitedescriptor.html)). Inside it you can add a section for custom properties `<custom>` http://maven.apache.org/plugins/maven-site-plugin/examples/sitedescriptor.html#Custom_Properties.

This could be an option, the thing is if we should support all the options from process-asciidoc mojo.

Also, so far I've only found a couple of example of how to recover custom properties and they use, either apache velocity scripts to directly generate the HTML, or some velocity api whose usage is not clear to me (https://github.com/andriusvelykis/reflow-maven-skin/blob/master/reflow-velocity-tools/src/main/java/lt/velykis/maven/skins/reflow/SkinConfigTool.java).
So, what I'm trying to say is that maybe getting the custom section to 'AsciidoctorParser' class won't be task for faint-hearted :smiley: Custom XML parsing is always an option, but I'm sure there must be a better way provided by the site-plugin tools.

mojavelinux:
One other possibility is to have a convention to set Maven properties with a special prefix to pass options to Asciidoctor.

```
<asciidoc-site-source-highlighter>coderay</asciidoc-site-source-highlighter>
```

or we could stuff them all into a single property.

```
<maven-site-asciidoc-attributes>source-highlighter=coderay toc=macro</maven-site-asciidoc-attributes>
```

[...]

PR #182, opened 2015-12-14
resolves #170, #180, #179, #168 & #160 enable configuration for site parser

### LINKED ISSUE #170: Unable the access Maven Site properties from Asciidoctor Maven Plugin
The Asciidoctor Maven Plugin does not have access to the attributes for Maven Site defined in the configuration section of the Maven Site plugin. Here is an example configuration:

              <plugin>
                    <groupId>org.apache.maven.plugins</groupId>
                    <artifactId>maven-site-plugin</artifactId>
                    <version>3.4</version>
                    <inherited>true</inherited>
                    <dependencies>
                        <dependency>
                            <groupId>org.apache.maven.wagon</groupId>
                            <artifactId>wagon-ssh</artifactId>
                            <version>1.0</version>
                        </dependency>
                        <dependency>
                            <groupId>org.asciidoctor</groupId>
                            <artifactId>asciidoctor-maven-plugin</artifactId>
                            <version>${asciidoctor.version}</version>
                        </dependency>
                        <dependency>
                            <groupId>lt.velykis.maven.skins</groupId>
                            <artifactId>reflow-velocity-tools</artifactId>
                            <version>${reflow-skin.version}</version>
                        </dependency>
                        
                        <dependency>
                            <groupId>org.apache.velocity</groupId>
                            <artifactId>velocity</artifactId>
                            <version>1.7</version>
                        </dependency>
                    </dependencies>
                    <configuration>
                        <attributes>
                            <ta>ddd</ta>
                        </attributes>
                    </configuration>
                </plugin>

I can't in an Asciidoctor document now use any of there attributes while it is possible to do it when using the Maven plugin only to render Asciidoctor documents.

mojavelinux:
- rewrite AsciidoctorParser (used by Maven Site Plugin) to make it more organized and extensible
- specify baseDir by default when invoking Asciidoctor API
- pass AsciiDoc configuration defined in Maven Site plugin to the Asciidoctor API
  - API options (e.g., templateDir)
  - Ruby options (e.g., requires)
  - AsciiDoc attributes (e.g., icons=font)
- enhance Maven Site integration test
  - add custom template
  - add include
  - add source highlighting
  - make validation code compatible with Java 6
- enable integration tests on Travis CI
- remove unused legacy site module
- fix license headers
- minor cleanups


src/main/java/org/asciidoctor/maven/site/AsciidoctorParser.java
```diff
@@ -31,31 +38,134 @@
...
+
+    protected String convertAsciiDoc(String source, OptionsBuilder options) {
+        return asciidoctor.convert(source, options);
+    }
+
+    private void requireLibrary(String require) {
+        if (!require.trim().isEmpty()) {
+            try {
+                // FIXME AsciidoctorJ should provide a public API for requiring paths in the Ruby runtime
+                RubyUtils.requireLibrary(JRubyRuntimeContext.get(), require.trim());
```
robertpanzer: Since Asciidoctorj 1.5.3 there's a method `Asciidoctor.requireLibrary(String)`

robertpanzer:
:+1: \o/

src/main/java/org/asciidoctor/maven/site/AsciidoctorParser.java
mojavelinux: I'll change this to a TODO since we haven't upgraded the AsciidoctorJ version yet. As soon as we do, we'll switch it over.

mojavelinux:
I agree, this is a major step forward. We just killed a whole army of bugs :)

jmbruel:
Congrats!👑

mnhock:
:+1:

mojavelinux:
I think this is ready to go!

abelsromero:
I'm really glad you could find a solution for this. I agree this is a major step and something worth of a 1.5.3 release.
I have a couple of questions:
* Should AsciidoctorMojo and AsciidoctorParser be refactored to use common methods?
* Should we use RubyUtils.requireLibrary or Asciidoctor.requireLibrary (from AsciidoctorJ)?

mojavelinux:
> Should we use RubyUtils.requireLibrary or Asciidoctor.requireLibrary (from AsciidoctorJ)?

We should use Asciidoctor.requireLibrary, but we have to do it after the fix for #183.

mojavelinux:
I added a note to #183.

mojavelinux:
> Should AsciidoctorMojo and AsciidoctorParser be refactored to use common methods?

Eventually, yes. Though AsciidoctorParser has more specialized requirements given that it cannot benefit from the type conversion provided by the `@parameter` injection.

What I really like now about the AsciidoctorParser class is that it would be easy to extend given that all the steps are partitioned into methods. In other words, you have hooks into the logic in order to customize the behavior. The main plugin should be this way too.

mojavelinux:
@abelsromero Let me know if you need more time to review. I'm ready to merge when you are.

robertpanzer:
Concerning `Asciidoctor.requireLibrary` or `RubyUtils.requireLibrary(JRubyRuntimeContext.get())`:

`JRubyRuntimeContext.get()` won't be available with AsciidoctorJ 1.6.0 as there is no longer one globale Ruby runtime.
Instead there is a new method `JRubyRuntimeContext.get(Asciidoctor)` that will return the Ruby runtime associated to the given Asciidoctor instance.

So better to use the first approach `Asciidoctor.requireLibrary(String)`.

mojavelinux:
:+1:

abelsromero:
> @abelsromero Let me know if you need more time to review. I'm ready to merge when you are.

GO!

mojavelinux:
:+1:

LightGuard:
Amazing the hoops you have to go through for this

PR #192, opened 2016-01-08
Add all of maven properties as asciidoctor properties. 

ptahchiev:
All of the maven properties are added as asciidoctor properties (replaced dots with underscore). This way we can reference them in our documentation like this:

```
{amazon_aws_sdk_version}
```


leif81:
I love this idea! :+1: 

Up to this point I've been mapping the Maven properties I care about individually with the plugin configuration.

e.g.

    <configuration>
        <attributes>
            <ProjectName>${project.name}</ProjectName>
            <ProjectVersion>${project.version}</ProjectVersion>
        </attributes>
    </configuration>

abelsromero:
I was doing some POCs for another issue and we will be injecting the MavenProject instance in AsciidoctorMojo too, so we could make it consistent for all mojos. Wdyt?

ptahchiev:
@abelsromero  I'm OK with this but again you will need to escape the properties to replace dots with underscore so that we can have `{project.properties.amazon_aws_sdk_version}`

abelsromero:
> you will need to escape the properties to replace dots with underscore

Sure, thanks. I'll just replicate your code. In the meantime do you think you could add some note about this in the README?
To tell the truth I don't know if  just add it into the "Passing POM properties" or update it and move it to "Tips & Tricks"

mojavelinux:
I also like this idea. Nice work.

However, this definitely needs to be something that requires an option to activate because it has the potential of causing unexpected/unwanted behavior for someone not anticipating it.

I also wonder whether we should add a prefix like pom- to avoid such conflicts.

LightGuard:
+1 for prefixing

On Friday, January 8, 2016, Dan Allen <notifications@github.com> wrote:

> I also like this idea. Nice work.
>
> However, this definitely needs to be something that requires an option to
> activate because it has the potential of causing unexpected/unwanted
> behavior for someone not anticipating it.
>
> I also wonder whether we should add a prefix like pom- to avoid such
> conflicts.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/asciidoctor/asciidoctor-maven-plugin/pull/192#issuecomment-170077206>
> .
>


-- 
Sent from Gmail Mobile

mojavelinux:
I also think we should use hyphen instead of underscore as the replacement character as that's more consistent with attribute names we use in the project.

LightGuard:
Works for me.

On Fri, Jan 8, 2016 at 11:11 AM, Dan Allen <notifications@github.com> wrote:

> I also think we should use hyphen instead of underscore as the replacement
> character as that's more consistent with attribute names we use in the
> project.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/asciidoctor/asciidoctor-maven-plugin/pull/192#issuecomment-170077826>
> .
>



-- 
Jason Porter
http://en.gravatar.com/lightguardjp

ptahchiev:
I'm also ok with all the above.

abelsromero:
Hi all,

Re-taking this PR in order to include it for the next release. The implementation for the other mojos and the site parser cannot be reused as I hoped.
Best thing, we can create an AttributesHelper class to hold common behaviour and keep it consistent. But we can do the refactor later on.
Also, to avoid confusions, this change (at least tested in the normal Mojo) allows to add as attributes exactly the elements enclosed in the <properties> sections (in the pom and profile). Things like `project.name` or `project.version` can be added but needs to be done explicitly.

So, to sum up, we can merge this, then add the same feature for mojos, release, and discuss separately which other properties would be interesting to include.

abelsromero:
I am thinking we can just merge this, it's been some time on hold for not real reason.

abelsromero:
I'll merge and replace with hyphens.

Issue #194, opened 2016-01-16
Integration tests fail with Maven 3.3.+ on Windows


abelsromero:
Since Maven 3.3.0 the command `mvn.bat` has been renamed to `mvn.cmd`.
This causes the `maven-invoker-plugin` (v1.8) to fail on Windows because it looks for a .bat file.

Newer version 2.0.0 supposedly fixes this issue ([MNG-5792](https://issues.apache.org/jira/browse/MNG-5792)) but test fail the same with no error message. Further investigation is required to find the cause.
Until this is fixed, the project will be stuck with using Maven 3.2.+ for the Appveyor tests.

Right now, the workarround for local environments is to copy & rename `mvn.bat` to `mvn.cmd` inside the Maven installation folder.

mojavelinux:
Why would they do this?

Maybe we could use that workaround in the integration test environment too. That would enable us to move forward.

abelsromero:
I just created the issue to ensure I did not forget this or in case someone else wanted to look into it.
Truth is after https://github.com/asciidoctor/asciidoctor-maven-plugin/pull/193 I want to check all versions because some are really old. For instance we are still using Maven API 2.2.1.

mojavelinux:
:+1:

abelsromero:
Run some test and found out that it was an issue with the execution rights of the maven command when using Sdkman (https://github.com/sdkman/sdkman-cli/issues/404) on Windows. Sdkman + MacOS works fine.
To be 100% sure, I created a branch in my repo with maven 3.3.3 and the new invoker plugin 2.0.0 and it works on Appveyour too.

I'll do the final PR later next week.

mojavelinux:
Excellent! Nice job on the investigative work.

abelsromero:
PR sent https://github.com/asciidoctor/asciidoctor-maven-plugin/pull/195

PR #195, opened 2016-01-19
Fixes issue #194

abelsromero:
A part from upgrading the version of the `maven-invoker-plugin` (2.0.0) and maven (3.3.9), I replaced the way Maven is installed.
Now it is download/installed directly without using chocolately. That way we don't need to wait for a package to be published in choco.
Thanks to @yegor256 and this article http://www.yegor256.com/2015/01/10/windows-appveyor-maven.html.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 target
 .settings
 .idea
+*.iml
 .classpath
 .project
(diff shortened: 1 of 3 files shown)


yegor256:
@abelsromero you're welcome :)

PR #197, opened 2016-01-26
Add parameter "failOnMissingSourceDirectory"

Stummi:
I want to use this plugin in a widely used parent pom, without breaking existing child projects which do not have the asciidoc source directory, so I decided to add a new option to change if the build fails in this case. I kept the current behavior as default. Please review and consider merging the PR

### DIFF
--- a/src/main/java/org/asciidoctor/maven/AsciidoctorMojo.java
+++ b/src/main/java/org/asciidoctor/maven/AsciidoctorMojo.java
@@ -133,6 +133,9 @@ public class AsciidoctorMojo extends AbstractMojo {
     @Parameter(property = AsciidoctorMaven.PREFIX + "attributeUndefined")
     protected String attributeUndefined = "drop-line";
 
+    @Parameter(property = AsciidoctorMaven.PREFIX + "failOnMissingSourceDirectory")
+    protected boolean failOnMissingSourceDirectory = true;
+
     @Override
     public void execute() throws MojoExecutionException, MojoFailureException {
         if (skip) {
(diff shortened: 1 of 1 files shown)


abelsromero:
I'm not sure if this may fit your case, but I think you can accomplish the same setting the plugin inside a profile with an `activation>file>exists` condition.
Something like:
```
    <profiles>
        <profile>
            <id>check-asciidoc-folder</id>
            <activation>
                <file>
                    <exists>src/docs/asciidoc</exists>
                </file>
            </activation>
            <build>
                <defaultGoal>process-resources</defaultGoal>
                <plugins>
                    <plugin>
                        <groupId>org.asciidoctor</groupId>
                        <artifactId>asciidoctor-maven-plugin</artifactId>
                  ....
```

mojavelinux:
I don't think we need the extra configuration option or the profile. The plugin shouldn't fail if the directory is empty. That would be like failing if there are no files to process. It should just move on. At worst it should log a debug message that it did no work because the directory does not exist.

Stummi:
I added it as option so it doesn't change current behavior of existing builds. However, if you think its no problem, just skipping the build without an option for it would be fine too

mojavelinux:
I appreciate that you made the effort to match the current behavior, but stepping back I think the current behavior is incorrect and we should just fix it at the root.

abelsromero:
> At worst it should log a debug message that it did no work because the directory does not exist.

I like that, at least that way you see if it was a configuration error.

@Stummi, will you do the honors to open an issue to keep track of this? Or if you preffer, it'll be a please to check/merge any related PR :wink:

Stummi:
Thanks. I simplified my changes and created a new PR as #198

PR #198, opened 2016-01-28
Skip processing on not existing source directory

Stummi:
Simplified from previous PR. See #197 for details


abelsromero:
For me it's OK. But would be perfect with a small test. wdyt?
That way we ensure that the behaviour is consistent with future changes.

This example shows how to to redirect the output to a String, I think it can be helpul. https://github.com/Stummi/asciidoctor-maven-plugin/blob/master/src/test/groovy/org/asciidoctor/maven/test/AsciidoctorMojoExtensionsTest.groovy#L271,L272

Issue #153, opened 2015-03-31
Remote include not working in the Maven plugin


mojavelinux:
A remote include from GitHub is not working when using the Maven plugin. It's resolving the include to an empty string.

Here's an example:

```
[indent=0]
....
include::https://raw.githubusercontent.com/cometd/cometd/master/pom.xml[tag=jetty-version]
....
```

This works when using both the Asciidoctor and AsciidoctorJ commandline tools. For some reason, it fails in the Maven plugin. We should include a test for this so we can ensure it doesn't break once we fix it.

mojavelinux:
This works in the Gradle plugin, so the problem seems to be with the Maven plugin or a conflict with Maven libraries. I remember @lordofthejars mentioning something about BouncyCastle perhaps?

paranoiabla:
It's probably because the HTML generated from Maven is different. This is because maven-site-plugin uses Doxia which sanitizes the produced html and removes a lot of the attributes on different tags (like class, style, etc..)

mojavelinux:
While that's true, this isn't related to that. The core functionality of Asciidoctor of including a file from a remote URL is failing in this case, but silently. I don't have time right now to debug, but I'm sure doing so would reveal what is happening...perhaps a swallowed exception.

LightGuard:
I'd be surprised if BouncyCastle had anything to do with this, it's for
cryptography. There may be a conflict somewhere, but we're not doing
anything with the asciidoc before we send it over. Maybe it's the security
we're running under?

On Tue, Mar 31, 2015 at 9:05 AM, Dan Allen <notifications@github.com> wrote:

> While that's true, this isn't related to that. The core functionality of
> Asciidoctor of including a file from a remote URL is failing in this case,
> but silently. I don't have time right now to debug, but I'm sure doing so
> would reveal what is happening...perhaps a swallowed exception.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/asciidoctor/asciidoctor-maven-plugin/issues/153#issuecomment-88124490>
> .
>



-- 
Jason Porter
http://en.gravatar.com/lightguardjp

mojavelinux:
I quickly tried a bunch of things. That's why I think getting it into the debugger is the next step because it's probably trying to tell us something, but it's getting swallowed.

mojavelinux:
[...]

PR #205, opened 2016-02-09
Reminder for issue #153

abelsromero:
As mentioned in isse #153 this test will ensure that we keep track of the issue when we finally fix it.

It has been tested that with jruby 9k the issue is resolved. When that happens (e.g., moving to Asciidoctorj 1.6.0) this tails will fail and will need to be updated.


abelsromero:
Note, that the issue will remain open untill fully fixed.

PR #209, opened 2016-02-20
fixes #206 - adds the option to set attributes through command line an…

### LINKED ISSUE #206: Adding support to define attribues using command line arguments
As mentioned in https://github.com/asciidoctor/asciidoctor-maven-plugin/issues/200 by @Mogztter, it would would be nice to be able to pass asciidoctor attributes using command line arguments like

````
mvn asciidoctor:process-asciidoc -Dasciidoctor.attributes="stylesheet=my-theme.css toc"
```

It is easy to implement, it only requires adding a new attribute and some maven anotations.

I have a working version https://github.com/abelsromero/asciidoctor-maven-plugin/blob/attributes_command/src/main/java/org/asciidoctor/maven/AsciidoctorMojo.java#L88-L89, and it works perfectly on all OS (Win, Lin, Mac). However, there is an open issue with the tests (https://issues.apache.org/jira/browse/MINVOKER-198).
To make things clear, only test fail, if I install the plugin in my local repo and use it, attributes can be passes as requested.

So, wdyt? Should we add it without an it-test for now?

abelsromero:
This PR reconfigures the attributes configuration in AsciidoctorMojo to allow setting them using command line options in the same as in asciidoctor.
This PR does not replace or change in any way how attributes are set using the standard XML.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -92,7 +92,7 @@ eruby:: defaults to erb, the version used in JRuby
 headerFooter:: defaults to `true`
 templateDir:: disabled by default, defaults to `null`
 templateEngine:: disabled by default
-sourceHighlighter:: enables and sets the source highlighter (currently `coderay` or `highlightjs` are supported)
+sourceHighlighter:: enables and sets the source highlighter (currently `coderay` or `highlight.js` are supported)
 attributes:: a `Map<String,Object>` of attributes to pass to Asciidoctor, defaults to `null`
 embedAssets:: Embedd the CSS file, etc into the output, defaults to `false`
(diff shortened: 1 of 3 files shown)


mojavelinux:
I'd just like to say, this is pretty darn cool.

One small request to the docs, though not directly related to this code change. Could you use "highlight.js" as the attribute value instead of "highlightjs"? We're trying to standardize that a bit and I think using the actually library name is the best way to go.

abelsromero:
Done! I also updated the tests and other references in the README.

mojavelinux:
Thanks!

Issue #196, opened 2016-01-25
Generated diagram images failed to be embedded in pdf


jinahya:
I'm generating diagrams. Images generated in `target/generated-docs-images`.
HTML successfully renders those generated images, but pdf fails.

```
asciidoctor: WARNING: image to embed not found or not readable: \
/path/to/src/main/asciidoc/images/diag-0eecdb8aae25da40f6cf41c33fce8cb1.png
```

I'm using following dependencies.

``` xml
        <dependencies>
          <dependency>
            <groupId>org.asciidoctor</groupId>
            <artifactId>asciidoctorj</artifactId>
            <version>1.5.4</version>
            <!--<version>1.6.0-alpha.1</version>-->
          </dependency>
          <dependency>
            <groupId>org.asciidoctor</groupId>
            <artifactId>asciidoctorj-diagram</artifactId>
            <version>1.3.1</version>
          </dependency>
          <dependency>
            <groupId>org.asciidoctor</groupId>
            <artifactId>asciidoctorj-pdf</artifactId>
            <version>1.5.0-alpha.11</version>
          </dependency>
          <!--          <dependency>
            <groupId>org.jruby</groupId>
            <artifactId>jruby-complete</artifactId>
            <version>9.0.4.0</version>
          </dependency>-->
        </dependencies>
```

mojavelinux:
When using Asciidoctor Diagram with the PDF converter, it's necessary to set the imagesdir to the output directory. That also means that you need to copy all the images there before the converter runs so it can find them all in one place.

This is a common problem when using Asciidoctor Diagram when the input and output directories are different (which is the case with any build tool). We really need to document this scenario clearly, but also find some way to make it simpler. It's a software design challenge.

mojavelinux:
The software design challenge is how to work with both input images and generated image across different folders referenced in different parts of the document. Right now, the only practical solution is to ensure all the images end up in one place before the converter runs.

abelsromero:
> We really need to document this scenario clearly

A part from adding some not in AsciidoctorJ manual, we can improve the current [example](https://github.com/asciidoctor/asciidoctor-maven-examples/tree/master/asciidoctor-diagram-example). Right now the examples is pretty simple.

jinahya:
I understand that problem is not simple.

With some curiosities, I tested.

* I put a static image on `src/main/asciidoc/images`
  * the static image copied to `target/generated-docs/images`
[...]

PR #193, opened 2016-01-11
Preview for #174 and #187 - Selecting and filtering sources and resources

abelsromero:
**PR description**
This PR is a preview of a solution for issues #174 and #187.
Both issues highlighted the limitations in terms of sources and resources management of the maven plugin.
Following the recomendation of Dan on isue #174
```
Agreed. We should align with the Gradle plugin on this one.
See https://github.com/asciidoctor/asciidoctor-gradle-plugin#usage (sources)
```
this PR includes the following changes:
(1). `sourceDirectory`, `sourceDocumentName` & `sourceDocumentExtension` disappear in favour of two new options sources & resources.
(2). `sources`: follows a the same resource pattern as the [maven-resource-plugin](https://maven.apache.org/plugins/maven-resources-plugin/examples/include-exclude.html). Basically it uses the same underlying API (https://maven.apache.org/shared/maven-filtering/). The difference here is that filtering/replacement options are not available and by default only AsciiDoc files are included.
Rendered files can be modified using custom include and exclude filters. This includes, adding more sources files with custom file extensions, adding multiple files, directories, etc.
(3). `resources`: follows the same resources pattern [...]. If not set, all resources (images, css, etc.) in the different sources are copied. This is different to how the gradle plugin works, but is consistent with how the current maven plugn works. Also, all source files defined using `include` conditions are excluded by default. In this case, the filtering (a.k.a. replacement) options provided by the maven-resource-plugin can be used, I would not advertise them.
All of this can be modified using custom include and exclude filters. For instance, if someone wants to copy AsciiDoc documents to the target folder, just create a <resource> tag like this one:
```
<resource>
    <directory>DIRECTORY</directory>
    <include>**/*.adoc</>
    <include>**/*.asciidoc</>
</resource>
```

(4). Options related to the output (format and directory) are kept 'global', that is, at <configuration> level as now. To some degree it makes sense to define them in the source element, but in my opinion this is just re-implementing maven executions inside the plugin. This includes options like `outputDirectory`, `baseDir`, `backend`, `preserveDirectories`, etc.
The criteria I followed was:
 * You need to manage several inputs? Now you can :)
 * You need to manage multiple outputs? You should use different executions. In the same way in gradle you would define a different task.
But eventually, this could be implemented if the community asks for it. But this will mean that resources will a property inside each source.

**Next steps**
* Le'ts discuss this and please let me know of corner cases to test. There's quite a lot of include/exclude manipulation and I suspect I may have slipt some special case.
[...]

PR #212, opened 2016-02-29
Added tip for combining Maven and Diagram

### LINKED ISSUE #207: How to use ditaa blocks with maven
When I add a [ditaa] block I get the following warning:

```asciidoctor: WARNING: index.adoc: line 17: invalid style for listing block: ditaa``

I use the asciidoctor maven plugin, asciidoctorj and asciidoctorjpdf in the following versions:

```
<asciidoctorVersion>1.5.3</asciidoctorVersion>
<asciidoctorjVersion>1.5.4</asciidoctorjVersion>
<asciidoctorjPdfVersion>1.5.0-alpha.11</asciidoctorjPdfVersion>
```

Do I miss something to be able to use the asciidoctor-diagram functionality with maven?

Thank you in advance

dashorst:
Closes #207 and #196

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -343,7 +343,7 @@ This is necessary since the `<configuration>` element is used to configure more 
 
 Here's an example that shows how to set options, attributes and ignore partial AsciiDoc files (i.e., files that begin with an underscore).
 
-[source,xml]
+[source,xml,subs=attributes+]
 .Maven site integration with Asciidoctor configuration
 ----
 <plugin>
@@ -371,7 +371,7 @@ Here's an example that shows how to set options, attributes and ignore partial A
         <dependency>
             <groupId>org.asciidoctor</groupId>
             <artifactId>asciidoctor-maven-plugin</artifactId>
-            <version>1.5.3</version>
+            <version>{release-version}</version>
         </dependency>
     </dependencies>
 </plugin>
@@ -460,7 +460,7 @@ When using Asciidoctor Diagram from the maven plugin you need a couple of things
 
 First you need to enable Asciidoctor Diagram by adding the gem to the maven plugin and instructing it to look for the generated images in the target folder:
 
-[source,xml]
+[source,xml,subs=attributes+]
 ----
 <plugin>
   <groupId>org.asciidoctor</groupId>
--- a/README.adoc
+++ b/README.adoc
@@ -453,3 +453,83 @@ NOTE: If I can figure out a good way to setup a Ruby testing environment I'll do
     ...
 </configuration>
 -----
+
+=== Using Asciidoctor Diagram
+
+When using Asciidoctor Diagram from the maven plugin you need a couple of things to set the build pipeline up correctly.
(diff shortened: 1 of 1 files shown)


abelsromero:
Thanks for the update :)
I have a couple of comments to add to it:
- I understand the meaning but I think that the first line mentioning `adding the gem` can be missleading. Adding the asciidoctorj-diagram jar is enough since it already wraps the gem.
- For the sake of completeness, I'd all both options. Using the maven resource plugin but also setting <imagesdir>. The maven resource is the best one, so we could leave the other as a small note for simple cases.
- The syntax used for the code examples is makdown-ish. The rest of the document (and being an Asciidoctor project one) uses AsciiDoc syntax.

wdyt?

dashorst:
If you remove the required gem from the configuration, asciidoctor-maven-plugin doesn't work with asciidoctor-diagram. I know because I just tried. Adding the gem is necessary.

Which setting are you talking about? You can't have static images and generated images without modifying both the `imagedir` attribute *and* adding the maven-resources-plugin.

abelsromero:
> If you remove the required gem from the configuration, asciidoctor-maven-plugin doesn't work with asciidoctor-diagram. I know because I just tried. Adding the gem is necessary.

Totally my mistake ^^'' I am so used to just copying and pasting some examples that I forget about that. The only thing you don't need is the gemPath, but you do need the require.

> Which setting are you talking about? You can't have static images and generated images without modifying both the imagedir attribute and adding the maven-resources-plugin.

Yes, thinking about it, the option to generate into the sources is not something I feel comfortable promoting.


BTW, since you're interested I had already worked on a prototype to integrate the maven-resources component into the plugin to be able to copy resources (https://github.com/asciidoctor/asciidoctor-maven-plugin/pull/193) before the rendering, thought the 'before' part was not done yet.
In the end I just canceled it because I was adding to many things in a single PR, but I planned to integrate only the feature to copy resources as soon as possible. Please, let me know what do you think and if you need it I can push to implement it this week, or feel free to add it yourself if like it.
That's the last thing I'd like to add before doing a new release.
I am saying this because I am not sure if your contribution will be maintained after the change and I feel bad :disappointed: about that.

dashorst:
Well, if you fix the plugin, then that's even better.

I think you might want to keep the text though, but change the wording. Some resources might need to be copied prior to the asciidoctor plugin running, so for this particular case it is not useful anymore with your new release, but other resources (movies?), perhaps from other folders might need to be copied prior to asciidoctor running.

abelsromero:
> I think you might want to keep the text though, but change the wording. Some resources might need to be copied prior to the asciidoctor plugin running, so for this particular case it is not useful anymore with your new release, but other resources (movies?), perhaps from other folders might need to be copied prior to asciidoctor running.

Apologies for taking so long to answer.
The answer is sure!, do you have some specific use case we can add with the new <resources> options? Also, keep in mind resources are copied before rendering.

PR #213, opened 2016-02-29
Add badge for Maven Central artifact

leif81:
A quick way for users to discover the latest deployed artifact version.


mojavelinux:
Looks good to me!

abelsromero:
I merged it but I now I see that clicking the image does not browse to maven central. Should we add a link to it?

leif81:
That's strange. Im pretty sure it was working in the PR. I wonder if the service had a regression just in the last day. I'll take a look later this morning.

PR #222, opened 2016-05-31
capacity to merge an existing theme and a project diff (in yaml) 

rmannibucau:
to have a custom patched theme for PDF backend

Notes:
- know the plugin doesnt integrate much with PDF there but the diff feature makes it quite nice and allow to share a big theme and change it by need directly in subprojects
- not sure if it requires PDF tests (just tested the diff logic)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -161,6 +161,11 @@
             <version>${doxia.version}</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>com.esotericsoftware.yamlbeans</groupId>
+            <artifactId>yamlbeans</artifactId>
+            <version>1.09</version>
+        </dependency>
     </dependencies>
 
     <build>
--- a/src/main/java/org/asciidoctor/maven/AsciidoctorMojo.java
+++ b/src/main/java/org/asciidoctor/maven/AsciidoctorMojo.java
@@ -12,18 +12,11 @@
 
 package org.asciidoctor.maven;
 
-import java.io.File;
-import java.io.FileFilter;
-import java.io.IOException;
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-
+import com.esotericsoftware.yamlbeans.YamlException;
+import com.esotericsoftware.yamlbeans.YamlReader;
+import com.esotericsoftware.yamlbeans.YamlWriter;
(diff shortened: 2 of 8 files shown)


mojavelinux:
I like what you are trying to accomplish, but I really don't think this belongs in the Maven plugin. This is really just a patch for https://github.com/asciidoctor/asciidoctor-pdf/issues/367, a fix which needs to happen in Asciidoctor PDF itself.

Maybe this could go in asciidoctorj-pdf, but even then I'm concerned about diverging behavior from what we implement in Asciidoctor PDF. I think the right solution is an `@import` directive just like CSS that we implement in Asciidoctor PDF. This is something the ThemeLoader should be handling.

rmannibucau:
Partly yes (the inheritance). The nice part about that code is also it handles the location for you keeping relative references working for the patched theme (for pictures etc).

ThemeLoader can certainly handles multiple paths relocation but have to admit 100 lines of code in the final artifact was a faster way to get unblocked that going through 3 releases.

abelsromero:
I am reviewing stuff. It is OK to I close this PR without merging?

r-tt:
@abelsromero have to admit i didnt test the ruby solution with mvn plugin (used a groovy script to prepare the theme in the meantime) but if the feature is done I'm happy to not pollute the mvn plugin with it, if not should be addressed somehow IMHO

rmannibucau:
@abelsromero have to admit i didnt test the ruby solution with mvn plugin (used a groovy script to prepare the theme in the meantime) but if the feature is done I'm happy to not pollute the mvn plugin with it, if not should be addressed somehow IMHO

mojavelinux:
I'm still not on board with this change. I don't think the Maven plugin should be working this low level. We need to get theme inheritance added to Asciidoctor PDF (which I 100% support) so that all the plugin has to worry about is where the theme file is.

rmannibucau:
@mojavelinux I'm +1 with that but also means the issue should stay open if I get/read it right (the issue was not about having the plugin doing it but having the feature, my PR was the fastest way to get it for me when I needed it)

mojavelinux:
I've opened an issue to track this request. Feel free to pile on the requirements there.

https://github.com/asciidoctor/asciidoctor-pdf/issues/571

mojavelinux:
I will certainly study this PR when implementing this feature. Let's just mark the pull request as "tracking" (aka upstream).

rmannibucau:
then no blocker to close or do whatever you want with this PR :), thanks Dan.

abelsromero:
I am fine with tracking 👍  
Thank you both

PR #223, opened 2016-06-06
Translate README into Chinese

diguage:


abelsromero:
Great! I'm thrilled to see how far the project is getting :)
But just one thing. I have no knowledge of Chinese so I cannot confirm the document nor mantain it, and I've seen how sometimes translations get a bit behind the main reference document.
So, to try to mitigate this how about we add a reference to the original Readme in the chinese page? Something with a comment like "This a translation of the main reference link_to_readme in cause of doubt of missing information please refer to it".
We can put in the same place as the chinese reference of the main README.
wdyt?

mojavelinux:
>  how about we add a reference to the original Readme in the chinese page? 

There is a link. I do like the idea of the disclaimer, though. We might also include in the disclaimer that if you notice an error in the translation, we welcome a fix for it.

> I have no knowledge of Chinese so I cannot confirm the document nor mantain it

Translations are a community effort. We have to trust that the Chinese-speaking segment of the community will validate it and be an active participant in maintaining the translation. The same is true for any translation. We'll never have translations unless we take this plunge. So we should take this plunge.

mojavelinux:
Please keep in mind that when this is merged, we should use the Squash + Merge option.

abelsromero:
> There is a link. I do like the idea of the disclaimer, though. We might also include in the disclaimer that if you notice an error in the translation, we welcome a fix for it.

@diguage wdyt about the disclamer? Let me know if it makes sense for you and I'll merge after that if you decide to include it:+1:

diguage:
@abelsromero That's OK. I will add the disclamer.

mojavelinux:
:+1:

Issue #217, opened 2016-04-13
Exclude custom resources by path


aleske:
Hi,

I'm generating documentation in multiple projects stored in SVN :-/
Each doc include the same header and footer file. These files are in a specific svn project.
In order to make this configuration work, I use svn:externals property.
Problem is .svn directory is created in the sourceDirectory.

Simple example
```
/doc/source/common/.svn <- here is the problem
/doc/source/common/header.adoc
/doc/source/myDoc.adoc
```

When copying resources, the FileFilter used only excludes asciidoc files. I don't want (and Jenkins is not allowed to do it anyway) svn files to be copied into the outputDirectory.

It would be a great idea to add a customizable FileFilter to permit exclusion like maven-site-plugin provides via 
```xml
<moduleExcludes>
    <asciidoc>**/_*.adoc</asciidoc>
</moduleExcludes>
```

Seems legit ?

abelsromero:
I feel this keeps biting from time to time...this has been in my head for some time.
Instead of just excluding, the idea (taken from the gradle plugin) is to add a <resources> section to filter which ones you want to copy. The resources section will work like in the maven-resources-plugin, so it should be familiar to most maven users and it allows to both define include patterns and exclude. What do you think?

You can find some preview implementation here (ignore the sources section, this will be for v1.6.+):
https://github.com/asciidoctor/asciidoctor-maven-plugin/pull/193

Let's use this conversation to gather all options and define the final solution. Do you think this can fully solve your case? any other scenario you can think off?

abelsromero:
PS: Cool image, brings back some childhood memories

abelsromero:
Hi @aleske, last PR should fix this issue. Before closing I'd like to check with you. 
A new element has been added to allow resource filtering following the same patterns as the maven-resources-plugin.

I have published a preview to test, the details are here:
https://bintray.com/abelsromero/maven/asciidoctor-maven-plugin/1.5.4-preview.2/


You can do something like this now:
```
<resources>
    <resource>
        <directory>${target_path}</directory> <!-- or just use . o filter across all files-->
        <exclude>**/**</exclude> <!-- does not copy anything -->
    </resource>
</resources>

aleske:
Oh ! I'm sorry for not having answered your previous message. I was busy as hell.

Anyway it seems perfect for me. I'm able to exclude svn files/folder and even the damned AsciidocFX generated folders (sometimes people commit somes).

Thanks !

[...]

Issue #187, opened 2016-01-06
sourceDocumentName with a document not in src/main/asciidoc


marcaurele:
Hi,

I have any issue with the change you did with the issue #58. If the path is relative to the expected source directory `src/main/asciidoc`, this directory **must** exist in order to process a file anywhere else in the project. My use case is a README.adoc in the root `${basedir}` I wish to process.

Setting the property to `<sourceDocumentName>../../../README.adoc</sourceDocumentName>` did not work because it fails while trying to copy the resources from `src/main/asciidoc` which does not exist.

Reading the comment of @mojavelinux on the issue #58 let me think that the plugin is not really design for such a use case. I just wanted to take up the subject to see if something could be done or not.

abelsromero:
Good news is that if you want to render the README in the root ${basedir} you can with the following options:
```
    <sourceDirectory>${basedir}</sourceDirectory>
    <sourceDocumentName>README.adoc</sourceDocumentName>
```

Bad news is that that case is not 100% covered. You can do that, but all folders and non-Asciidoc files will be copied in the output directory. This incudes 'src/main/**', `src/test/**` and so on...

Ideally we want to be able to set filters to control what gets rendered/copied (see https://github.com/asciidoctor/asciidoctor-maven-plugin/issues/174), but right now the option is to render the file, copy the folders and just use some other maven plugin to remove what you don't need.

marcaurele:
Yep, I tested with scenario too but since it copies everything, it's not an option for me.

My use case can be one too if you have the README.adoc rendered by github and wish to generate the same file in your final archive. So for me it makes sense to find a solution for it, with the possibility to list multiple files (README.adoc, INSTALL.adoc...).

The first thing to fix is to not raise an error when trying to copy the assets if the `src/main/asciddoc` does not exist.

abelsromero:
> The first thing to fix is to not raise an error when trying to copy the assets if the src/main/asciddoc does not exist.

Can you please post your configuration? I don't get that error and maybe you have found another related bug.

marcaurele:
Ok, I won't be able to do it before Monday.

abelsromero:
No worries :smile:  Post it when you can.

marcaurele:
Originally it was splitted into 2 pom files (parent, child). I put everything in one, it should not make a difference.
```xml
[...]

PR #225, opened 2016-06-19
Adds support to define resources directories with include/exclude filters

abelsromero:
This PR fixes issues:
- #187 and #217: being able to exclude/filter the resources that are copied into the output directory.
- #196: using asciidoctor-diagram with asciidoctor-pdf requires copying the resources before the AsciiDoc documents are rendered. ->_I added this change because it was just moving 2 lines of code_

Following the approach in [PR 192](https://github.com/asciidoctor/asciidoctor-maven-plugin/pull/193) I have added the `maven-filtering` library to allow defining sets of resources like in this example:
````
<resources>
    <resource>
        <!-- add some images and exclude SVN files -->
        <directory>DIRECTORY</directory>
        <include>**/*.jpg</include>
        <include>**/*.gif</include>
        <exclude>**/.svn</exclude>
    </resource>
    <resource>
        <!-- add some sources that are linked in the docs -->
        <directory>DIRECTORY_2</directory>
        <include>**/*.java</include>
        <exclude>**/.class</exclude>
    </resource>
    <resource>
        <!-- exclude all -->
        <directory>DIRECTORY_3</directory>
        <exclude>**/**</exclude>
    </resource>
</resources
```

This breaks no compatibility with current versions, but has some drawbacks:
- Additional log messages from the 'maven-filtering' component appear showing the list of resources being copied.
- Since resources are copied before rendering, if something fails during rendering resources are kept in the target folder.


Note:
If this is accepted, there are some TODOs for the near future:
- README: section about `Using Asciidoctor Diagram` can be simplified to only require setting `imagesdir` to the output folder.
- maven examples: after release, I'd like to update the maven diagram example to include flat images also and show how both (generated by diagram and flat) can co-exists.
- lots of refactoring...I tried to keep changes to minimum for now, but at some point I'd like to improve the tests structure and `AsciidoctorMojo` class. The later has grown a lot and is a bit messy.

### DIFF
--- a/src/main/java/org/asciidoctor/maven/AsciidoctorMojo.java
+++ b/src/main/java/org/asciidoctor/maven/AsciidoctorMojo.java
@@ -12,38 +12,31 @@
 
 package org.asciidoctor.maven;
 
-import java.io.File;
-import java.io.FileFilter;
-import java.io.IOException;
-import java.lang.reflect.InvocationTargetException;
-import java.lang.reflect.Method;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-
 import org.apache.commons.io.FileUtils;
+import org.apache.maven.execution.MavenSession;
+import org.apache.maven.model.Resource;
 import org.apache.maven.plugin.AbstractMojo;
 import org.apache.maven.plugin.MojoExecutionException;
 import org.apache.maven.plugin.MojoFailureException;
+import org.apache.maven.plugins.annotations.Component;
 import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
-import org.asciidoctor.AbstractDirectoryWalker;
-import org.asciidoctor.AsciiDocDirectoryWalker;
-import org.asciidoctor.Asciidoctor;
-import org.asciidoctor.Attributes;
-import org.asciidoctor.AttributesBuilder;
-import org.asciidoctor.DirectoryWalker;
-import org.asciidoctor.Options;
-import org.asciidoctor.OptionsBuilder;
-import org.asciidoctor.SafeMode;
+import org.apache.maven.project.MavenProject;
+import org.apache.maven.shared.filtering.MavenFilteringException;
+import org.apache.maven.shared.filtering.MavenResourcesExecution;
+import org.apache.maven.shared.filtering.MavenResourcesFiltering;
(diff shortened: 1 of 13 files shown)


src/main/java/org/asciidoctor/maven/AsciidoctorMojo.java
```diff
@@ -184,7 +204,11 @@ public void execute() throws MojoExecutionException, MojoFailureException {
                 throw new MojoExecutionException(e.getMessage(), e);
             }
         }
-        
+
+        // Copy output resources
+        repareResources();
```
mojavelinux: Did you mean to type "prepare" instead of "repare"?

src/main/java/org/asciidoctor/maven/AsciidoctorMojo.java
```diff
@@ -242,10 +319,9 @@ private void setDestinationPaths(OptionsBuilder optionsBuilder, final File sourc
 
     protected Asciidoctor getAsciidoctorInstance(String gemPath) throws MojoExecutionException {
         Asciidoctor asciidoctor = null;
-        if (gemPath == null) {
-             asciidoctor = Asciidoctor.Factory.create();
-        }
-        else {
+        if (gemPath == null || gemPath.isEmpty()) {
```
mojavelinux: I think a value of empty string is supposed to be passed to `create()`. This ensures that Asciidoctor is isolated.

mojavelinux:
I'm really looking forward to this change.

src/main/java/org/asciidoctor/maven/AsciidoctorMojo.java
abelsromero: I totally believe you but I want to understand this. Wasn't `null` the trick to really isolate it? I recall this has been changed in 1.6.0 to inject a null when using 'create()'.
So, what's the difference between create(), create(null) and create("")?

mojavelinux: If I remember correctly, there is a different. The empty string is the default, and that does need to be passed to create() for Asciidoctor 1.5.x. You are correct that 1.6.0 fixes this. What I don't remember is how to make the value null. I think if you put an empty tag (`<gemPath/>`), that makes the value null. A null value means "don't isolate" in this context.

abelsromero: Thanks :smile:

abelsromero:
If no one has anything else to add I'll squash/merge it later.

PR #226, opened 2016-06-25
Resource filtering fix

abelsromero:
This PR includes the fixes discussed in issue #187.
The required libraries to enable filtering have been added and now it's possible to use it. Hovewer, this is not exposed in the documentation.

Here is the list of changes in this PR:

* Updated README to align with `maven-resource-plugin` configuration and explain the supported options. Note that using only <include> without <includes> works also, I guess it's my gradle side that preffers it.
*  Added integration test for property filtering (a.k.a. replacing). Even if it's not officially supported, that way we have an example if someone wants to use it and asks about it.
* Fixed an issue: when defining custom _file extensions_ in `sourceDocumentExtension`, these were also copied. Now they are excluded.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -73,15 +73,15 @@ There are several configuration options that the asciidoctor-maven-plugin uses, 
 sourceDirectory:: defaults to `${basedir}/src/main/asciidoc`
 sourceDocumentName:: an override to process a single source file; defaults to all files in `${sourceDirectory}`
 sourceDocumentExtensions:: (named `extensions` in v1.5.3 and below) a `List<String>` of non-standard file extensions to render. Currently ad, adoc, and asciidoc will be rendered by default
-resources:: list of resource files to copy to the output directory (e.g., images, css). The configuration follows the same patterns as the `maven-resources-plugin`
+resources:: list of resource files to copy to the output directory (e.g., images, css). The configuration follows the same patterns as the `maven-resources-plugin`. If not set, all resources inside `sourceDirectory` are copied
 +
 [source, xml]
 ----
     <resources>
         <resource>
-            <!-- (Mandatory) Directory to copy from. By default uses maven's ${baseDir} -->
+            <!-- (Mandatory) Directory to copy from. Paths are relative to maven's ${baseDir} -->
             <directory>DIRECTORY</directory>
-            <!-- (Optional) Directory to copy to. By default uses the option `sourceDirectory` -->
+            <!-- (Optional) Directory to copy to. By default uses the option `outputDirectory` -->
             <targetPath>OUTPUT_DIR</targetPath>
             <!-- (Optional) NOTE: SVN, GIT and other version control files are excluded by default, there's no need to add them -->
             <excludes>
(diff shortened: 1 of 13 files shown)


abelsromero:
comments?

README.adoc
```diff
@@ -73,19 +73,28 @@ There are several configuration options that the asciidoctor-maven-plugin uses,
...
 [source, xml]
 ----
     <resources>
         <resource>
+            <!-- (Mandatory) Directory to copy from. By default uses maven's ${baseDir} -->
             <directory>DIRECTORY</directory>
-            <include>**/*.jpg</include>
-            <include>**/*.gif</include>
-            <exclude>**/.svn</exclude>
+            <!-- (Optional) Directory to copy to. By default uses the option `sourceDirectory` -->
```
mojavelinux: I don't think this comment is correct. The resources aren't copied to the source directory but rather the output directory, correct?

src/it/resource-filtering/pom.xml
```diff
@@ -0,0 +1,49 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<groupId>org.asciidoctor</groupId>
+	<artifactId>test</artifactId>
+	<version>1.0-SNAPSHOT</version>
+
+	<name>Run Asciidoctor Article to Html</name>
```
mojavelinux: Instead of "Run", I would recommend the word "Convert"

README.adoc
abelsromero: Reading it again I think that the overall section is not clear. I propose the next text.
- Replaced first line telling that if resources are not set `all resources inside `sourceDirectory` are copied`
- Corrected descriptions for <directory> and <targetPath>
```
resources:: list of resource files to copy to the output directory (e.g., images, css). The configuration follows the same patterns as the `maven-resources-plugin`. If not set, all resources inside `sourceDirectory` are copied
+
[source, xml]
----
    <resources>
        <resource>
            <!-- (Mandatory) Directory to copy from. Paths are relative to maven's ${baseDir} -->
            <directory>DIRECTORY</directory>
            <!-- (Optional) Directory to copy to. By default uses the option `outputDirectory` -->
            <targetPath>OUTPUT_DIR</targetPath>
            <!-- (Optional) NOTE: SVN, GIT and other version control files are excluded by default, there's no need to add them -->
            <excludes>
                <exclude>**/.txt</exclude>
            </excludes>
            <!-- (Optional) If not set, includes all files but default exceptions mentioned -->
            <includes>
                <include>**/*.jpg</include>
                <include>**/*.gif</include>
            </includes>
        </resource>
        <resource>
                ...
    <resources>
----
```

src/it/resource-filtering/pom.xml
abelsromero: All integration tests should be reviewed with that an other improvements.
Wdyt about opening a separate issue for that?

abelsromero:
Sorry to ask, but what do you think of the proposals?

abelsromero:
@mojavelinux I am a bit confused, I see the 'thumbs' up in the issue, but since it does not show the dat I am not sure if this has been review.

Is it good to merge?

mojavelinux:
I didn't think about the fact that the reaction icon doesn't show the date. When I'm giving a :+1: to say "proceed with merge" I'll use a comment instead. In this case, I was giving the thumbs up to merge.

abelsromero:
Good to know I did not screw up :smile:

mojavelinux:
Not at all. You're doing fantastic, both on the code and communicating.
We're so lucky to have you at the lead!

PR #227, opened 2016-07-09
Upgrading Netty

Sanne:
Upgrading Netty and some maven plugins.

I'd like to get this working on Java 9; there is still much more to do but this is a start.


abelsromero:
Thanks a lot, that's been a pending task for a long time.
Just to double check, I also run the tests in a Java6 HotSpot machine and everything works fine :+1:  

The current plan is to drop support for Java6 with the 1.6.0 version.

mojavelinux:
:+1:

Sanne:
thanks!

PR #252, opened 2016-07-18
fixes #228 - No source code highlighting

### LINKED ISSUE #228: No source code highlighting
I have an AsciiDoc file that contains some source code; also, its header contains the line `:source-highlighter: highlight.js`.

When I generate HTML via `asciidoctor document.adoc`, the resulting HTML file correctly highlights the source code.

However, when I generate HTML for the same file via the maven plugin, the source code has no highlighting at all. I tried adding the configuration attribute `<source-highlighter>highlight.js</source-highlighter>` in `pom.xml` (just because the Readme told me to), but it didn't make any difference.

The difference between the generated files is that the AsciiDoctor build contains the following lines, which enable syntax highlighting; the Maven build does not.

```xml
<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/8.9.1/styles/github.min.css">
<script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/8.9.1/highlight.min.js"></script>
<script>hljs.initHighlighting()</script>
```

How can I instruct the Maven plugin to include these lines the way AsciiDoctor does?

abelsromero:
The initialization of `sourceHighlighter` [here](https://github.com/asciidoctor/asciidoctor-maven-plugin/blob/master/src/main/java/org/asciidoctor/maven/AsciidoctorMojo.java#L111) with `""`, caused that the value in the document was overridden precisely `""`.
Just setting it to null fixes the issue.


mojavelinux:
:+1:

PR #255, opened 2016-07-21
Invoke JavaExtensionRegistry by reflection to take care of binary inc…

robertpanzer:
…ompatibilities with AsciidoctorJ 1.6.0

Related to asciidoctor/asciidoctorj#491

To also support AsciidoctorJ 1.6.0 it will be necessary to invoke JavaExtensionsRegistry via Reflection.

### DIFF
--- a/src/main/java/org/asciidoctor/maven/extensions/AsciidoctorJExtensionRegistry.java
+++ b/src/main/java/org/asciidoctor/maven/extensions/AsciidoctorJExtensionRegistry.java
@@ -1,5 +1,7 @@
 package org.asciidoctor.maven.extensions;
 
+import org.apache.maven.plugin.MojoExecutionException;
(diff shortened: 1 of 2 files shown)


abelsromero:
:+1:

mojavelinux:
Nice work!

I think we should mark that this should be changed back to a regular invocation once we move to 1.6.x version of the plugin. Version 1.6.x of the plugin will require version 1.6.x of AsciidoctorJ...and thus we will be back in binary alignment. wdyt?

robertpanzer:
Like this? :-)
https://github.com/asciidoctor/asciidoctor-maven-plugin/pull/255/files#diff-3c89526d74c66dc30084df700cd6c102R55

mojavelinux:
:+1:

PR #266, opened 2016-09-13
fixes #264 - adds support for sourcemap, catalog_assets and template_cache options

### LINKED ISSUE #264: Adding sourcemap option when converting documents
I'm working on an extension (based on treeprocessor) that is supposed to be invoked via asciidoctor-maven-plugin. The extension should provide line number, asciidoctor file name where error is occurred. 
If understand it right, sourcemap option should be specified to get line numbers from nodes. Looking on the maven plugin sources, it looks like there is no possibility to pass that option to asciidoctorj API. 

While it's not possible to pass sourcemap option via the maven plugin configuration attributes, Could you please suggest if it is possible pass sourcemap option via Ruby code that could be injected by using "requires" attribute?

abelsromero:
The PR also adds better explanations for other options.
No tests are included since validation of those options require Asciidoctorj 1.6.0. When this is released, tests using Java extensions can be added.
Options are set using property names due to them not being available in OptionsBuilder.

I have validated the code using an external project that uses AsciidoctorJ 1.6.0-alpha.3.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -118,9 +118,12 @@ backend:: defaults to `docbook`
 doctype:: defaults to `null` (which trigger's Asciidoctor's default of `article`)
 eruby:: defaults to erb, the version used in JRuby
 headerFooter:: defaults to `true`
-templateDir:: disabled by default, defaults to `null`
-templateEngine:: disabled by default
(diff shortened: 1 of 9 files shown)


abelsromero:
closing, need to fix merges

Issue #269, opened 2016-09-20
Exclude docinfo files by default


ge0ffrey:
With for example foo-docinfo.html and foo-docinfo-footer.html we can add a snippet of html code in asciidoctor to the generated html. The plugin shouldn't copy these as resources in the output directory by default.

mojavelinux:
I agree!

The following files should be ignored:

* docinfo.html
* docinfo-header.html
* docinfo-footer.html
* *-docinfo.html
* *-docinfo-header.html
* *-docinfo-footer.html

(docinfo-header is reserved for future use, but not yet enabled)

ge0ffrey:
I tried workaround around it with this config in the plugin, but that doesn't work for some reason:

              <goals>
                <goal>process-asciidoc</goal>
              </goals>
              <configuration>
                <resources>
                  <resource>
                    <directory>src/main/asciidoc</directory>
                    <excludes>
                      <exclude>**/docinfo.html</exclude>
                      <exclude>**/docinfo-header.html</exclude>
                      <exclude>**/docinfo-footer.html</exclude>
                      <exclude>**/*-docinfo.html</exclude>
                      <exclude>**/*-docinfo-header.html</exclude>
                      <exclude>**/*-docinfo-footer.html</exclude>
                    </excludes>
                  </resource>
                </resources>

I'll investigate, to see if I find a workaround and maybe send in a PR to fix this issue upstream.

abelsromero:
Resource filtering has not been published yet, if you want to use it you'll need to download the sources and compile them locally.
With the current published version there's no way to accomplish what you need. Only work-arround I can think of is deleting the files after they are generated using another plugin.

ge0ffrey:
PR send: https://github.com/asciidoctor/asciidoctor-maven-plugin/pull/270

PR #270, opened 2016-09-21
269 Exclude docinfo files by default

ge0ffrey:
Fixes https://github.com/asciidoctor/asciidoctor-maven-plugin/issues/269

### DIFF
--- a/src/main/java/org/asciidoctor/maven/io/AsciidoctorFileScanner.java
+++ b/src/main/java/org/asciidoctor/maven/io/AsciidoctorFileScanner.java
@@ -26,7 +26,13 @@ public class AsciidoctorFileScanner {
             "**/docinfo-footer.html",
             "**/*-docinfo.html",
             "**/*-docinfo-header.html",
-            "**/*-docinfo-footer.html"};
+            "**/*-docinfo-footer.html",
+            "**/docinfo.xml",
+            "**/docinfo-header.xml",
+            "**/docinfo-footer.xml",
+            "**/*-docinfo.xml",
+            "**/*-docinfo-header.xml",
+            "**/*-docinfo-footer.xml"};
 
     private BuildContext buildContext;
 
--- a/src/main/java/org/asciidoctor/maven/io/AsciidoctorFileScanner.java
+++ b/src/main/java/org/asciidoctor/maven/io/AsciidoctorFileScanner.java
@@ -14,8 +14,19 @@ import java.util.*;
 public class AsciidoctorFileScanner {
 
     public static String[] DEFAULT_FILE_EXTENSIONS = {"**/*.adoc", "**/*.ad", "**/*.asc","**/*.asciidoc"};
-    // Files and directories beginning with underscore are ignored
-    public static String[] IGNORED_FOLDERS_AND_FILES = {"**/_*.*", "**/_*", "**/_*/**/*.*"};
+
+    public static String[] IGNORED_FOLDERS_AND_FILES = {
+            // Files and directories beginning with underscore are ignored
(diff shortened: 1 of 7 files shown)


abelsromero:
Just one thing, could you add some test?
I know in this case it's not 100% necessary, but it helps a lot in the future to ensure behavior stays consistent if we change something.

ge0ffrey:
Got a pointer to a test I can copy-paste-adjust?

abelsromero:
Sure!
You can use this method https://github.com/ge0ffrey/asciidoctor-maven-plugin/blob/master/src/test/groovy/org/asciidoctor/maven/test/AsciidoctorMojoTest.groovy#L32-L59

You'll need to:
- remove the resources filter
- add the attribute configuration, IIRC `mojo.attributes['docinfo1']= 'true'`
- add a docInfo.html file to the src/test/resources directory
- validate that the file is not found in the target and that the content is added to the text. The test has examples about that.

Also, to ensure (to some degree) tests isolation, you can add some timestamp to the output like this `File outputDir = new File("target/asciidoctor-output/${System.currentTimeMillis()}")`

mojavelinux:
Should we exclude all docinfo file extensions as well? In other words, when converting to DocBook, is the plugin currently copying the .xml files?

abelsromero:
The PR shows the only filters that are currently applied (_ prefix). So docinfo XML files are copied.
The thing is that if we want to be 100% accurate we should filter only the appropriate files based on the backend and docinfo attribute (docinfo, docinfo1 and docinfo3), and even take into consideration the `<docname>`. Is not hard, but requires some work.
I am basing my assumptions on these: 
http://asciidoctor.org/docs/user-manual/#docinfo-attributes-and-file-names and
http://asciidoctor.org/docs/user-manual/#footer-docinfo-files, we need additional logic.

mojavelinux:
While you are indeed correct, I personally feel like that would be overkill. The term "docinfo" in general should be considered reserved and files that include it (particularly with the patters we've put forth) should not expected to be copied. I'm not opposed to a more fine-grained approach...I'm just saying it isn't strictly necessary.

ge0ffrey:
I agree that taking backend and docinfo into account would be better, but imo it shouldn't prevent merging this PR as is. The lack of a test is indeed a problem, I 'll update the pr with a test tomorrow.

Any plans to do a release of this plugin?

abelsromero:
> While you are indeed correct, I personally feel like that would be overkill.

I am glad we all agree on that. So filtering all options (html and xml) should be enough.


> Any plans to do a release of this plugin?

@ge0ffrey, about the release, there are a couple PR that would be nice to include but they aren't mandatory. The only blocking thing for me is the release of AsciidoctorJ 1.5.4.2 which includes a patch to fix an issue with extension https://github.com/asciidoctor/asciidoctor-maven-plugin/pull/258. Without it, some plugin tests fail.

ge0ffrey:
PR updated with unit test and exclusion of docinfo.xml variants. Should be ready to merge, do note that this is my first Groovy unit test :)

abelsromero:
> do note that this is my first Groovy unit test :)

I have no changes to propose. It can surely be optimized, but as for me, I prefer readability in tests over fancy one-liners.

I'll leave it open while we wait to see about the AsciidoctorJ release (https://github.com/asciidoctor/asciidoctorj/pull/498), just in case someone else wants to add anything.

ge0ffrey:
Thanks :)

abelsromero:
Thanks to you. As a general note, if you see I don't answer/merge in a couple of days, feel free to ping me on the thread or twitter.

Issue #244, opened 2016-07-17
Maven Configuration in appveyor.yml does not use the cache


khmarbaise:
Based on the docs maven should be configured to use the local build cache to prevent downloading of all artifacts for every build. 
`-Dmaven.repo.local=C:\maven`

khmarbaise:
The cache is cleaned every time the `appveyor.yml` changes. sounds like a very good strategy (http://www.appveyor.com/docs/build-cache).

mojavelinux:
What about the `C:\Users\appveyor\.m2` folder? Isn't that where the cache is located?

mojavelinux:
Oh, I see...you are configuring the location of the cache. It seems like it would be better to use the cache under the user's folder and just mark that as a cache folder.

khmarbaise:
If this is the correct location it would be simpler to configure this in `appveyor.yml` and remove the `-Dmaven.repo.local=C:\maven` from the maven calls.

mojavelinux:
:+1:

abelsromero:
Closing with PR https://github.com/asciidoctor/asciidoctor-maven-plugin/pull/272

PR #272, opened 2016-09-28
Added cache control and Java6 build to appveyor

abelsromero:
This PR includes:
- Cache for maven installation folder and repository. Each cache adds a depedency file (https://www.appveyor.com/docs/build-cache/#addchange-dependency), so that if that file changes the cache gets invalidated.
- Support for Java6 builds. Since officialy until asciidoctor 1.6.0 Java6 is supported and now Appveyor offers it, I added it. To make the build work for Java6, it uses maven 3.2.5 instead. Higher versions fail because they are compiled for Java7. `RefreshTest` hangs, could not reproduce, so I disabled it for Java6.

Finally, Maven wrapper has not been included to support the Java6.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,18 +2,27 @@ version: '{build}'
 skip_tags: true
 clone_depth: 10
 environment:
+  MAVEN_VERSION: 3.3.9
   matrix:
+    - JAVA_HOME: C:\Program Files\Java\jdk1.6.0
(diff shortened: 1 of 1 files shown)


abelsromero:
If there're no comments I'll merge later this afternoon. Since there's already an issue to setup maven wrapper, I'll also close the PR #244

Issue #278, opened 2016-11-29
Attributes from pom.xml not substituted in PDF (sometimes)


LaurianeDPX:
I use Asciidoc Maven plugin to generate a PDF. I have this at the beginning of the document:

```
= {document-title} 
{organization}
v{project-version}, {localdate}
:numbered:
:toc: macro
```

`document-title`, `organization` and `project-version` come from my pom.xml:

```
<configuration>
    <attributes>
        <document-title>${project.name}</document-title>
        <project-version>${project.version}</project-version>
        <organization>${project.organization.name}</organization>
    </attributes>
</configuration>
```

My problem is that `project-version` is not substituted by its value, it is just blank.

The interesting thing is when I use this:

```
v{project-version}, 2016-11-29
```
or this:
```
v{project-version}
```
`project-version` is substituted !

Any idea how to solve this ? Thanks !

Additional information:
- `asciidoctor-maven-plugin` version 1.5.3
- `asciidoctorj-pdf` version 1.5.0-alpha-11
- `jruby-complete` version 9.0.5.0
- `asciidoctorj` version 1.5.4

And the doc type is `book`.

abelsromero:
I reproduced the issue using directly ascidoctor 1.5.5 (ruby) in the shell. /cc @mojavelinux 
```
asciidoctor -b html5 -a document-title="AsciiDoc to HTML Maven Example" -a project-version="1.0.0-SNAPSHOT" -a organization="my super org" demo.adoc
```
`project-version` get replaced in text, but not in the header.

In my case however, I can only show the version in the header when it's alone. No other option works.
Also, if instead of `project-version` I use `version-label` it work,s but I don't think it is legal to replace that internal attribute.

mojavelinux:
It may be that substitutions aren't performed in revision line (I'd need to check). But it shouldn't matter since there is a far simpler way to do this. Just leave off the revision line and set these attributes:

```
<configuration>
    <attributes>
        <document-title>${project.name}</document-title>
        <revnumber>${project.version}</revnumber>
        <revdate>${maven.build.date}</revdate>
        <organization>${project.organization.name}</organization>
    </attributes>
</configuration>
```

LaurianeDPX:
Your solution works like a charm @mojavelinux and thanks for your very quick answers.
I had to use `${maven.build.timestamp}` instead of `${maven.build.date}` though.

Thank you !

mojavelinux:
:+1:

[...]

PR #279, opened 2016-11-30
Improved Tips & Tricks section with an example of header attributes

abelsromero:
This PR contains:
· As a closure for https://github.com/asciidoctor/asciidoctor-maven-plugin/issues/278 I added the resolution into the tips & trick section of the README document.
· Added brief text descriptions of the different Tips&Tricks.
· Removed Tip note about use of `revnumber` in `Passing POM properties`. I consider that the new tip is more elegant and having both can create some confusion.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -211,15 +211,6 @@ The custom AsciiDoc property can then be used in the document like this:
 
  The latest version of the project is {project-version}.
 
-[NOTE]
-====
-If you want to have the project version as the revision number of the document, use this construct:
-
- :revnumber: {project-version}
-
-This will make the version number appear in the header and footer of the output.
-====
-
 ==== Setting boolean values
 
 Boolean attributes in asciidoctor, such as `sectnums`, `linkcss` or `copycss` can be set with a value of `true` and unset with a value of `false`.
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -484,3 +479,21 @@ NOTE: If I can figure out a good way to setup a Ruby testing environment I'll do
...
+Automatically add version details to header and footer to all documents.
+
+[source, xml]
+-----
+<configuration>
+    ...
+    <attributes>
+        ...
+        <revnumber>${project.version}</revnumber>
+        <revdate>${maven.build.date}</revdate>
```
mojavelinux: This should actually be maven.build.timestamp...or we should explain how to get Maven to expose the build date. Maybe even suggest integrating with the Maven git plugin. Otherwise, people are going to ask by maven.build.date doesn't work.

abelsromero: I'd rather keep the tip simple and void other plugins integration.
Instead I fixed the use of `date` by `timestamp` property and also added how to deifne the date format.

mojavelinux:
Good. I didn't realize `maven.build.timestamp` was set by default, but it is since Maven 2.1. So it's all good.

Issue #291, opened 2017-03-31
PDF generation doesn't respect "resources" configuration attribute


ge0ffrey:
Using the "resources" element in the plugin configuration, we can grab images from 2 directories. We use this to include shared images and a book-specific images. This works perfectly for the html output.

However for the PDF output, configured like this:

```
              <goals>
                <goal>process-asciidoc</goal>
              </goals>
              <configuration>
                <backend>pdf</backend>
                <resources>
                  <resource>
                    <directory>src/main/asciidoc</directory>
                  </resource>
                  <resource>
                    <directory>${project.basedir}/../shared-kie-docs/src/main/asciidoc</directory>
                  </resource>
                </resources>
                <sourceDocumentName>index.adoc</sourceDocumentName>
                <outputDirectory>${project.build.directory}/generated-docs/pdf</outputDirectory>
              </configuration>
```

The images end up broken in the pdf and the maven console shows messages like these for images not residing in src/main/asciidoc:
```
  asciidoctor: WARNING: image to embed not found or not readable: /home/.../kie-docs/docs/optaplanner-wb-es-docs/src/main/asciidoc/KieServer/kie-server-simple-architecture.png
asciidoctor: WARNING: image to embed not found or not readable: /home/.../kie-docs/docs/optaplanner-wb-es-docs/src/main/asciidoc/KieServer/kie-server-architecture.png
```

abelsromero:
That's kind of expected. Maybe we should add a warning in the README to avoid confusions.
The plugin just copies the resources to the target(s), but it does not modify any configuration, for instance, in this case it does not modify the `imagesdir` attribute.
That affects backends that embed images or other resources from source directory like PDF.

In order to solve this, you can setup the `imagesdir` accordingly in every document section when rendering to pdf like this:
```
= Title
:imagesdir: images

This is just a document to test images includes.

== First Section

This images is in `images` directory.

image::asciidoctor-icon.jpg[]

== Second Sections
// Replaces imagesdir value from this point on when using PDF backend
ifdef::backend-pdf[]
:imagesdir: more-images
endif::[]

This images isn't.
image::ruby.png[]
```
But **beware**, because this affects the document, so once imagesdir is modified the new value remains.

Alternatively, you could try to build an extension that adjustes the imagesdir dinamically looking for images in diferent directories. But I think the option above is more manteinable.

mojavelinux:
I think we need to take a step back and focus on "That's kind of expected".

[...]

PR #292, opened 2017-04-01
related to #291 - added note regarding configuration of multiple resources

abelsromero:
In the meantime a final implementation is provided, a note has been added to README to clarify that some backends need additional configurations when multiple resources are set.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -101,8 +101,10 @@ resources:: list of resource files to copy to the output directory (e.g., images
 +
 [NOTE]
 ====
-Backends that embed resources, need to locate them at the time of conversion.
-For example, in the case of pdf and images, all reosurces should be aggregated under a single directory, and `imagesdir` attribute should be set to point to it.
+Converters that embed resources such as images into the output document need to be able to locate those resources at conversion time.
+For example, when generating a PDF (or HTML with the `data-uri` attribute set), all images need to be aggregated under a common root (i.e., image catalog).
+The `imagesdir` attribute should be overridden to point to that folder.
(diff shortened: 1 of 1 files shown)


abelsromero:
AppVeyour caches have been giving problems lately :/ I'll try to find some time this week to check it.

README.adoc
```diff
@@ -99,6 +99,12 @@ sourceDocumentName:: an override to process a single source file; defaults to al
 sourceDocumentExtensions:: (named `extensions` in v1.5.3 and below) a `List<String>` of non-standard file extensions to render. Currently ad, adoc, and asciidoc will be rendered by default
 resources:: list of resource files to copy to the output directory (e.g., images, css). The configuration follows the same patterns as the `maven-resources-plugin`. If not set, all resources inside `sourceDirectory` are copied
 +
+[NOTE]
+====
+Backends that embed resources, need to locate them at the time of conversion.
+For example, in the case of pdf and images, all reosurces should be aggregated under a single directory, and `imagesdir` attribute should be set to point to it.
+====
```
mojavelinux: I'd like to propose the following revision:

```
Converters that embed resources such as images into the output document need to be able to locate those resources at conversion time.
For example, when generating a PDF (or HTML with the `data-uri` attribute set), all images need to be aggregated under a common root (i.e., image catalog).
The `imagesdir` attribute should be overridden to point to that folder.
When converting to HTML, images must be copied to the output location so that the browser can resolve those images when the user views the page.
```

mojavelinux:
:+1:

PR #293, opened 2017-04-07
upgrade Jruby dependency to 9K

guyuqi:
We are facing the below issue when we are building Apache HBase on AArch64：　
[ERROR] Failed to execute goal org.asciidoctor:asciidoctor-maven-plugin:1.5.2.1:process-asciidoc (output-pdf) on project hbase: Execution output-pdf of goal org.asciidoctor:asciidoctor-maven-plugin:1.5.2.1:process-asciidoc failed: (NotImplementedError) fstat unimplemented unsupported or native support failed to load ->

This error took place because asciidoctor-maven-plugin used an old version of Jruby (1.7.26) which didn’t support AArch64  (due to the fact that Jffi 1.2.12 which Jruby 1.7.26 depends on has no native AArch64 libraries).

Is it ok for upgrade jruby version to 9K to support AArch64?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -70,7 +70,7 @@
         <asciidoctorj.version>1.5.5</asciidoctorj.version>
         <maven.coveralls.plugin.version>4.3.0</maven.coveralls.plugin.version>
         <maven.jacoco.plugin.version>0.7.2.201409121644</maven.jacoco.plugin.version>
-        <jruby.version>1.7.26</jruby.version>
+        <jruby.version>9.1.8.0</jruby.version>
         <maven.plugin.plugin.version>3.4</maven.plugin.plugin.version>
         <maven.plugin.api.version>2.0</maven.plugin.api.version>
         <maven.project.version>2.2.1</maven.project.version>


abelsromero:
There's no need to change the plugin configuration. You can change the version on your build through the plugin's dependency section. Here is an example: https://github.com/asciidoctor/asciidoctor-maven-examples/blob/master/asciidoctor-pdf-example/pom.xml#L33-L37.
Same goes for the AsciidoctorJ version.
The plugin provides the minimum compatible version for simplicity sake, but it's not mandatory to use that specific version.

Btw, I noticed the error says version 1.5.2.1, you may be interested in upgrading to the last version 1.5.5.

guyuqi:
Hi @abelsromero ,
Really thanks for your tips. But unfortunately, the last version 1.5.5 also use the jruby 1.7.26 which do not support AARCH64.

When I build hbase that includes its dependency "asciidoctor-maven-plugin" on AArch64,  I must download "asciidoctor-maven-plugin" source code , change the jruby version and install it(mvn clean install ) on my local maven base.  

Could you please tell me why there's no need to upgrade the jruby in asciidoctor-maven-plugin? Thanks!

abelsromero:
We do plan to mode to jRuby 9K in future versions, but not now because we still want to support Java 6.
I know it's pretty outdated, but it was a milestone defined a long time ago and we want to stick to it.
Said that, I am afraid my explanation was not clear enough. You do not need to change the dependency and install locally, you can set the version in your own pom and the plugin will use it instead of the default one.
Checking the HBase repo, I see you're already adding asciidoctorj-pdf [here](https://github.com/apache/hbase/blob/master/pom.xml#L1062), you just need to add the the jruby 9k dependency below. If you point me to how to send a PR to the Apache Git I can do the changes.

guyuqi:
It's cool.  I'll do the changes in HBase. Really thanks for the info.

PR #297, opened 2017-04-17
Improved message on non-existing source directory

obfischer:
Added information about the expected source directory to the message logged if the expected directory does not exist. This could be helpful for the user.


abelsromero:
:+1

abelsromero:
If no one objects I'll merge in a few hours.

PR #301, opened 2017-07-17
added outputFile option to override the generated filename

tisoft:
The asciidoctor CLI allows to override the output filename with the -o, --out-file arguments. This change allows to do the same with the maven plugin.

This allows for example to include the project version in the filename:
```xml
<outputFile>my_lovely_documentation_${project.version}.pdf</outputFile>
```

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -130,6 +130,7 @@ When converting to HTML, images must be copied to the output location so that th
     <resources>
 ----
 outputDirectory:: defaults to `${project.build.directory}/generated-docs`
+outputFile:: defaults to `null`, can be used to override the name of the generated output file. This is useful for backends, that create a single file, e.g. the pdf backend. All output will be redirected to the same file, the same way as the `-o, --out-file=OUT_FILE` option from the `asciidoctor` CLI command.
 baseDir:: (not Maven's basedir) enables to set the root path for resources (e.g. included files), defaults to `${sourceDirectory}`
 skip:: set this to `true` to bypass generation, defaults to `false`
 preserveDirectories:: enables to specify whether the documents should be rendered in the same folder structure as in the source directory or not, defaults to `false`.
--- a/src/main/java/org/asciidoctor/maven/AsciidoctorMojo.java
+++ b/src/main/java/org/asciidoctor/maven/AsciidoctorMojo.java
@@ -538,6 +538,14 @@ public class AsciidoctorMojo extends AbstractMojo {
         this.outputDirectory = outputDirectory;
     }
 
+    public File getOutputFile() {
+        return outputFile;
+    }
+
+    public void setOutputFile(File outputFile) {
+        this.outputFile = outputFile;
+    }
+
     public String getBackend() {
         return backend;
     }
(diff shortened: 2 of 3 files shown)


abelsromero:
I see no problem with the PR, but I feel we need to warn that files may be overwritten.
Since all documents share the same name, all documents rendered to the same target directory are overwritten by the last one.

We could add a warning message in the code, or just state this in the documentation. wdyt?

Btw, AppVeyor passed OK.

abelsromero:
Also, it's worth mentioning that the path in `outputFile` is nicely appended to the source paths when using `outputDirectory` with and without `preserveDirectories`. :+1:

tisoft:
I would add it to the documentation. It does the same as the -o, --out-file=OUT_FILE command line option from the asciidoc CLI command.

abelsromero:
Perfect! Do you want to it yourself to this PR? Or else, I can find some time tomorrow to do it.

tisoft:
I have added it to the README.adoc. I hope the wording is ok. 😏

I wasn't able to do the 汉语 translation, though. 😁

README.adoc
```diff
@@ -130,6 +130,7 @@ When converting to HTML, images must be copied to the output location so that th
     <resources>
 ----
 outputDirectory:: defaults to `${project.build.directory}/generated-docs`
+outputFile:: defaults to `null`, can be used to override the name of the generated output file. This is useful for backends, that create a single file, e.g. the pdf backend. All output will be redirected to the same file, the same way as the `-o, --out-file=OUT_FILE` option from the `asciidoc` CLI command.
```
abelsromero: You mean `asciidoctor` command right?

abelsromero:
I have one minor comment. I typo a guess :P 
oh! and thanks a lot for adding a test.

README.adoc
tisoft: Yes, have fixed it. But I just checked, the asciidoc command supports the same parameter, so I wasn't wrong 😁

abelsromero: Thanks. My comment is just cause I want to avoid controversy with the python implementation.

mojavelinux:
> Since all documents share the same name, all documents rendered to the same target directory are overwritten by the last one.

I would say that it should only be used if there's exactly one input file. An alternative would be to use it, but warn if there's more than one input file. Either way, I agree it shouldn't silently work if there are multiple input files.

abelsromero:
I was happy with the mention in the README, but we can add a explicit warning in the build process.
To do it properly we need to take into consideration `preserveDirectories` as well, store the target paths and warn whenever a repeated one appears. That way the users know exactly what happened.
Also, on top of the warning, we could skip the rendering of the file, to 'fail' faster.

mojavelinux:
I was just thinking that if there is more than one file being processed (which the plugin should know), then it's a warning. This feature is essentially mutually exclusive with processing more than one file.

abelsromero:
But if files are in different directories and `preserveDirectories` is enabled, there's no collision. I admit it's odd, but still this causes no issue. Maybe I am overthinking?

PR #307, opened 2017-08-28
Upgrade asciidoctorj to 1.5.6

### LINKED ISSUE #257: Upgrade asciidoctorj to the latest version

tisoft:
Fixes #257


abelsromero:
Thanks 👍 I guess we should start thinking about a release soon.

PR #309, opened 2017-09-11
Fix/upgrade TravisCI to trusty

abelsromero:
*This is not a final PR, open for discussion*

This PR updates the TravisCI configuration to the new images based on Ubuntu Trusty, set as default since September (https://blog.travis-ci.com/2017-08-31-trusty-as-default-status),
The changes from TravisCI are:
- Default support only for Java8 (oracle & openjdk) and OpenJDK 7.
- Default maven version is 3.5.0

This means that:
- We no longer test against oracle 7
- **We should drop support for Java6. This I decided on my own and I'd like to hear feedback.** If we all agree, I'll update the PR removing the current hack to install it.
Thing is AsciidoctorJ no longer supports Java6, so I don't see any reason to support it on the Maven plugin. If we want to support it, some nasty hacks needs to be added to the script, similar to the ones applied to AppVeyour. Nothing impossible, but honestly, it seems a waste of time to me.

Btw, same changes should be applied to AsciidoctorJ build, but I found this issue https://github.com/jruby-gradle/jruby-gradle-plugin/issues/305.
Forcing TravisCI to use `precise` instead of trusty fixes it, but I don't think we should do that. Support for precise is not guaranteed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ matrix:
         apt:
           packages:
             - oracle-java7-installer
-      jdk: oraclejdk78
+      jdk: oraclejdk7
     - addons:
         apt:
           packages:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,20 @@ before_script:
   - export MAVEN_SKIP_RC=true
 jdk:
   - oraclejdk8
-  - oraclejdk7
+  - openjdk8
   - openjdk7
-  - openjdk6
+matrix:
+  include:
+    - addons:
+        apt:
+          packages:
+            - oracle-java7-installer
+      jdk: oraclejdk78
+    - addons:
+        apt:
+          packages:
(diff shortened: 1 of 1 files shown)


mojavelinux:
I agree this is the right move forward. For builds that pre-dated the trusty rollout, it's necessary to specify. But it just seems like a good practice to specify it anyway because then you know what image is being used to run the CI build.

I also agree we should drop Java 6. We should probably formalize this further and say that we support the last two major versions. That's a pretty standard policy. That way, when Java 9 is final, we should drop Java 7. (If that seems too soon, we can say the last 3 versions).

abelsromero:
It failed for some `random` error, relaunching the job fixed it. If there are no more comments I'll merge and proceed to apply changes to Asciidoctorj, in that case the only working config currently is using `precise`.

PR #311, opened 2017-09-16
Set java version to 7 in maven-compiler-plugin

abelsromero:


abelsromero:
Not sure `project.execution.environment` is used at all, but I don't want to introduce potential issues right now.

PR #313, opened 2017-09-16
fixes #303 - adds warning every time rendered file is overwritten by …

### LINKED ISSUE #303: Show warning when `outputFile` is enabled and more than one file is processed
As mentioned in https://github.com/asciidoctor/asciidoctor-maven-plugin/pull/301.
A warning should be shown when more than one file is processed while `outputFile` is enabled.

A possible option is simply log a warning message when more than one file is detected during input processing.
Another options is checking when a generated file is overriden by another during generation, and show a message only then.

abelsromero:
As title suggests, adds the warning message below every time it detects that it will overwrite an existing file 
 `"Duplicated destination found: overwriting file: " + destinationPath.getAbsolutePath()` 
The file get overwritten any way to keep current behavior, but it can be very easily changed to skip the rendering, opinions?
Also not sure if we should use Warn or Err level. ERR seems to harsh, but I also think that overwriting files is a sign of a bad build and should be fixes, same as a code bug.

The idea was to address the use of `outputFile` option easily, but I found the issue may happen in other cases. For example, rendering AsciiDoc docs that are in different source paths into the same target directory (which is the default behauvior). That's why I decided to go for a full feature that checks all files.

Now, before rendering each file, it will check if the target path has already been used. I wanted to pre-process all target paths in a similar way as how resources are prepared before copying them. But the current code makes it very complicated and inefficient. Doing that would allow for some optimization in the future.
I really want to do a major refactor for 1.5.7. I am not afraid to admit that current code is a good example of `spaghetti code`

PR also includes:
- code typo fix
- Note on usage of absolute paths in `outputFile` + test

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -130,7 +130,7 @@ When converting to HTML, images must be copied to the output location so that th
     <resources>
 ----
 outputDirectory:: defaults to `${project.build.directory}/generated-docs`
-outputFile:: defaults to `null`, can be used to override the name of the generated output file. This is useful for backends, that create a single file, e.g. the pdf backend. All output will be redirected to the same file, the same way as the `-o, --out-file=OUT_FILE` option from the `asciidoctor` CLI command.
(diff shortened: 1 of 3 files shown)


mojavelinux:
I support this approach. I think a warning is sufficient. We don't really have to worry about destruction of files since this is a build directory. The configuration can easily be corrected and the build run again to restore the files that were overwritten.

Thank you for writing such thorough tests. That's the most important part of this change.

PR #316, opened 2017-09-27
Removed old bundler directory

abelsromero:
Removes old `bundler` folder. I suspect it has been there since the first releases that used the gem directly doing nothing.

PS: now that we are free of Java6 support. I'll do some maintenance in the repo, expect some silly small PRs: updating plugins, refactor tests, and other stuff.


mojavelinux:
I don't think the plugin ever used the gem directly, but it might have been there for local testing of development releases of the gem. I agree it doesn't need to be there.

PR #334, opened 2018-01-20
fixes #330 - Reused attributes treatment for site goal

### LINKED ISSUE #330: `site` goal - self-closed attributes (eg. <sectnums/>) do not work
All attributes need to be set in a full form.
But some do not have value and should be able to be set using self-closed form.
For example `sectnums` needs to be set as `<sectnums>true</sectnums>`.

abelsromero:
Just created and utility class to reuse the fix already present in the main mojo.

I have left the code unchanged but I have two notes:
1. I think that "true", "false" and Boolean could be treated together with AsciidoctorJ's `Attributes.toAsciidoctorFlag`. Imho, it makes the code more readable.
2. I don't see why this logic could not be included directly in AsciidoctorJ AttributesBuilder. Attributes class should not do anything, but the Builder purposes is to help and make configuration easier right?
/c @robertpanzer

### DIFF
--- a/src/main/java/org/asciidoctor/maven/AsciidoctorHelper.java
+++ b/src/main/java/org/asciidoctor/maven/AsciidoctorHelper.java
@@ -1,0 +1,59 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the &quot;License&quot;);
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


abelsromero:
I'll merge tomorrow if no one has anything to mention

PR #336, opened 2018-01-21
Improved site mojo documentation in README

abelsromero:
Added:
* Clarification on AsciiDoc documents and resources locations
* Detailed list of supported options in the site: with its limitations and quirks


popebear:
The updated doc is clear - thanks very much for looking into that.

PR #337, opened 2018-02-11
Migrate plugin to Maven 3+ only

### LINKED ISSUE #233: Migrate plugin to Maven 3+ only
Currently the plugin is using maven-plugin-api version 2.0 whereas the maven-project is 2.2.1 which is not consistent. Furthermore I think it's time to migrate to Maven 3+ only.
[Maven 2.2.1 has End Of Life](http://maven.apache.org/maven-2.x-eol.html) for more than two years in the meantime.
I thinks that should also in a major version upgrade to make clear that this is not compatible with Maven 2.X anymore ...

slachiewicz:
Required Maven version updated to 3.0.4

Fixes #233

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -325,6 +325,62 @@
                 </lifecycleMappingMetadata>
               </configuration>
             </plugin>
+            <!-- default plugins for build, if not defined in parent-pom -->
+            <plugin>
+              <groupId>org.apache.maven.plugins</groupId>
+              <artifactId>maven-source-plugin</artifactId>
+              <version>3.0.1</version>
+            </plugin>
+            <plugin>
+              <groupId>org.apache.maven.plugins</groupId>
+              <artifactId>maven-javadoc-plugin</artifactId>
+              <version>3.0.0</version>
+            </plugin>
+            <plugin>
+              <groupId>org.apache.maven.plugins</groupId>
+              <artifactId>maven-clean-plugin</artifactId>
+              <version>3.0.0</version>
+            </plugin>
+            <plugin>
+              <groupId>org.apache.maven.plugins</groupId>
+              <artifactId>maven-deploy-plugin</artifactId>
+              <version>2.8.2</version>
+            </plugin>
+            <plugin>
+              <groupId>org.apache.maven.plugins</groupId>
+              <artifactId>maven-install-plugin</artifactId>
+              <version>2.5.2</version>
+            </plugin>
+            <plugin>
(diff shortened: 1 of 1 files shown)


abelsromero:
Don't know why, but it running very slowly on Windows, AppVeyor failed because of that and it also happens in my machine. Maybe we should upgrade each dependency one by one and see what is the source of it :/

Other that that, I think Maven ships with commons io 2.5, we should probably stick to that version to play safe.

slachiewicz:
First run was slow propably due to depenedency downloads - could You check if the next builds are also slow?
I've created update to a Maven cache and of course I can also create separate patches to update the dependences.

> Cache 'C:\Users\appveyor\.m2' - Invalidated as some dependencies (pom.xml) have changed

abelsromero:
I don't think the issue comes form caches. If you compare the execution time of tests form this
https://ci.appveyor.com/project/asciidoctor/asciidoctor-maven-plugin/build/234/job/s1vae6o2o8dus08l
to 
https://ci.appveyor.com/project/asciidoctor/asciidoctor-maven-plugin/build/233/job/bk2f3frr9pxwxth1

Some tests took more than 30 minutes to run, I look for "Time elapsed" text.

slachiewicz:
Sure, that makes sense - I will split it into smaller, independent and testable parts.

abelsromero:
Thanks a lot for looking into it. Tbh, I am really puzzled by this...but as always, the issue is in Windows.

abelsromero:
I narrowed it down to the surefire plugin. All versions starting from 2.19 make the doc render run very slow.
Version 2.18.1 run nicely.
Other than that, just removed the plugins I think we don't need. Here is my POC https://github.com/abelsromero/asciidoctor-maven-plugin/tree/poc/upgrade_to_maven_v3

slachiewicz:
Looks good, we can close my pull now.

slachiewicz:
The documentation includes information about required version of maven 3.2.1, and in prereq says 3.0.5. Which one is right?

abelsromero:
> Looks good, we can close my pull now.

Since you opened the issue, I was hoping we could reuse your branch. 

> The documentation includes information about required version of maven 3.2.1, and in prereq says 3.0.5. Which one is right?

That's probably for the site plugin. I guess we should upgrade to 3.2.1.

abelsromero:
Done https://github.com/asciidoctor/asciidoctor-maven-plugin/pull/345

PR #339, opened 2018-02-11
Use Maven 3.5.2 for AppVeyor build and update .m2 cache

### LINKED ISSUE #338: Update Maven for AppVeyor and update .m2 cache settings
Every update to pom invalidate the m2 cache, so in order not to slow down a pull request validation - invalidate only when appveyor.yml file update.

slachiewicz:
Every update to pom invalidate the m2 cache, so in order not to slow down the compilation - invalidate only when appveyor.yml file update.

Fixes #338

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,11 +2,11 @@ version: '{build}'
 skip_tags: true
 clone_depth: 10
 environment:
-  MAVEN_VERSION: 3.5.0
+  MAVEN_VERSION: 3.5.4
   matrix:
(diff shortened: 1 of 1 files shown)


abelsromero:
While I find some time to test the dependencies on a Windows machine we cna work on this.
I am not sure that the chace should be linked to appveyour. That would mean that if we change the project dependencies but not the build, the build will use the cache right?

slachiewicz:
Yes, cache will stay for longer and will not be invalidated with every change to pom.xml (previously this was a trigger)

PR #341, opened 2018-02-11
Remove dependency to oss-parent and copy to local pom.xml

slachiewicz:
First - remove only parent-pom.
Next, i will add selected plugin updates from #337 

cc: @khmarbaise could You help with validation?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -50,21 +50,7 @@
         <developerConnection>scm:git:git@github.com:asciidoctor/asciidoctor-maven-plugin.git</developerConnection>
     </scm>
 
-    <repositories>
-        <repository>
-            <id>sonatype-nexus-snapshots</id>
-            <name>Sonatype Nexus Snapshots</name>
-            <url>https://oss.sonatype.org/content/repositories/snapshots</url>
-            <releases>
-                <enabled>false</enabled>
-            </releases>
-            <snapshots>
-              <enabled>true</enabled>
-            </snapshots>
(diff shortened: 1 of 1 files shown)


abelsromero:
Looked into [this](https://maven.apache.org/guides/mini/guide-central-repository-upload.html) and I think we can totally remove the parent. After all we are publishing through jcenter.

slachiewicz:
Yes, we can remove parent-pom, but we should include all that was defined here. 
I try to compare effective pom (_mvn help:effective-pom_) before, and after patch and looks equal.
Later you can remove a reduntant entries

abelsromero:
> Yes, we can remove parent-pom, but we should include all that was defined here.

I totally agree with that to be safe. But I think we can go a bit further ;)

The oss-parent `maven-release-plugin` adds the `sonatype-oss-release` profile which overwrites `source`, `javadoc` and `gpg` plugins. But those are already defined, and honestly the options the oss build sets seem to be outdated to me.

I am ok with going ahead and seeing if we can simplify later. But wdyt, do we take our change now?

mojavelinux:
I'm completely in support of this change. There was a time when that parent made sense. That time has passed. So this is really just residue from an old setup.

slachiewicz:
Ok, i've removed redundant sonatype-oss-release profile and snapshot repository. 
It was good for me to check how deployment to jcenter works - thx :) 
After squash - I think all should be ready now for merge.

PR #342, opened 2018-02-18
Set default encoding to UTF-8 for resource copy plugin

slachiewicz:
Set default encoding for mock Mojo


slachiewicz:
Removes the warning at build on Windows:
_[warn] Using platform encoding (Cp1252 actually) to copy filtered resources, i.e. build is platform dependent!_

abelsromero:
:+1:

PR #350, opened 2018-06-05
Added an option to replace the constant "@{project.basedir}" with the current modules base directory

chrisdutz:
Especially when generating the output of the diagram asciidoctor plugin into the correct directory, all output is always output into the root modules site directory. This breaks things for child modules. 

Unfortunately there doesn't seem to be a maven way to provide the local modules output directory without modifying the plugin. 

This Pull-Request enables the plugin to replace the constant string "@{project.basedir}" with the current modules base directory. Allowing to provide things like this:

`    <plugin>
          <groupId>org.apache.maven.plugins</groupId>
          <artifactId>maven-site-plugin</artifactId>
          <version>3.7.1</version>
          <configuration>
            
            <generateReports>true</generateReports>
            <generateSitemap>true</generateSitemap>
            <relativizeDecorationLinks>false</relativizeDecorationLinks>
            <locales>en</locales>
            <inputEncoding>${project.build.sourceEncoding}</inputEncoding>
            <outputEncoding>${project.reporting.outputencoding}</outputEncoding>
            <asciidoc>
              <attributes>
                <source-highlighter>prettify</source-highlighter>
                <imagesoutdir>@{project.basedir}/target/site/img</imagesoutdir>
              </attributes>
              <requires>
                <require>asciidoctor-diagram</require>
              </requires>
            </asciidoc>
          </configuration>
       ...
`
It's not a pretty solution, but it works for my use-cases. Unfortunately Doxia doesn't seem to provide a clean way to handle this.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -73,7 +73,8 @@
         <plexus.component.metadata.version>1.5.5</plexus.component.metadata.version>
         <netty.version>4.0.38.Final</netty.version>
         <commons.io.version>2.5</commons.io.version>
-        <doxia.version>1.7</doxia.version>
+        <!-- Updated to 1.8 in order to have RenderingContext.getBasedirRelativePath() -->
+        <doxia.version>1.8</doxia.version>
         <gmaven.version>1.5</gmaven.version>
     </properties>
 
@@ -143,6 +144,11 @@
         <dependency>
             <groupId>org.apache.maven.doxia</groupId>
             <artifactId>doxia-core</artifactId>
+            <version>${doxia.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.maven.doxia</groupId>
+            <artifactId>doxia-site-renderer</artifactId>
             <version>${doxia.version}</version>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -452,8 +452,6 @@
                             <preBuildHookScript>setup</preBuildHookScript>
                             <postBuildHookScript>validate</postBuildHookScript>
                             <showErrors>true</showErrors>
-                          <!-- Try to find out why the ci builds are failing -->
-                          <streamLogs>true</streamLogs>
                             <debug>true</debug>
                             <!-- <localRepositoryPath>${project.build.directory}/local-repo</localRepositoryPath>-->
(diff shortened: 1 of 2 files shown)


chrisdutz:
Ok ... seems there are problems when running my version with older maven-site-plugin versions. But I still think the patch is not pretty ... so I'll fix that if you guys agree that you want it ;-)

chrisdutz:
Ok ... so I'm giving up on trying to understand why the build is failing ... eventually someone who knows how to track this problem down could provide some feedback.

abelsromero:
About the PR, at first glance it does not seem bad, but need to check it more. I hope to have some time next week, sorry :%

The build error seems to be with an integrated test. I image this does not fail on your machine right?

Btw, if you have issues with a maven plugin locally you can use `mvnDebug` to debug a maven build same way to debug into a remote server.

chrisdutz:
Yeah ... I usually just run my Maven build in IntelliJ ... then it's like debugging any other application. But you are correct, I'm not getting the errors at all.

So did you find some time to check?

chrisdutz:
Ping?

abelsromero:
Haven't been able to have a look at it yet, sorry.

abelsromero:
I checkout your branch and the site integration test fails with:
```
Caused by: java.lang.NoClassDefFoundError: org/apache/maven/doxia/siterenderer/DocumentContent
    at java.lang.Class.getDeclaredConstructors0 (Native Method)
    at java.lang.Class.privateGetDeclaredConstructors (Class.java:2671)
    at java.lang.Class.getDeclaredConstructors (Class.java:2020)
```
If you run the tests locally with `mvn verify -Prun-its`, you can find the reports in under `target/it`

abelsromero:
I installed the branch locally and run this test repo (https://github.com/abelsromero/asciidoctor-maven-multimodule) with it and, if I undestood correctly the issues still happen.
What I see is that:
- Some resources are generated at the root.
- diagram images are generated next to the source (in project-2 they are generated on target)
¿Is the test repo right or I am missing something?

chrisdutz:
Hi ... seems your changes making the MavenProject available in the doxiaRenderer solved my problem, so that soltution is a lot prettier than mine ... so I'll close this PR. Thanks

PR #359, opened 2018-07-17
Added TSLv1.2 flag to appveyor build

abelsromero:
Since June 18 TSL 1.0 & 1.1 are not supported by Maven Central and that causes the Java7 build to fail in AppVeyour.
Added `-Dhttps.protocols=TLSv1.2` to force use of TSL v1.2 and a note to remind us to remove it once we don't support Java7.


abelsromero:
PS: maven-site integration test fails, but will be fixed in another PR

Issue #326, opened 2017-11-22
Is this plugin thread safe?


shalugin:
I want to use parallel compilation in maven.

I have the following warning:
```

[WARNING] *****************************************************************
[WARNING] * Your build is requesting parallel execution, but project      *
[WARNING] * contains the following plugin(s) that have goals not marked   *
[WARNING] * as @threadSafe to support parallel building.                  *
[WARNING] * While this /may/ work fine, please look for plugin updates    *
[WARNING] * and/or request plugins be made thread-safe.                   *
[WARNING] * If reporting an issue, report it against the plugin in        *
[WARNING] * question, not against maven-core                              *
[WARNING] *****************************************************************
[WARNING] The following plugins are not marked @threadSafe in MapStruct Documentation:
[WARNING] org.asciidoctor:asciidoctor-maven-plugin:1.5.6
[WARNING] Enable debug to see more precisely which goals are not marked @threadSafe.
[WARNING] *****************************************************************
```

Is `asciidoctor-maven-plugin` thread safe?

abelsromero:
Good question! We should test it and do some research to be 100% sure, but I am pretty sure the plugin is,
We comply to everything stated here (except for third libraryes, need to check): https://cwiki.apache.org/confluence/display/MAVEN/Parallel+builds+in+Maven+3
Afaik AsciidoctorJ is thread-safe (cc @robertpanzer ) and we don't keep anything in static variables.

The way I understand the documentation, Maven parallelises at module level not at Mojo level. And each execution of the mojo uses a new class instance of it in a single JVM.
Note that the warning comes because of the metadata descriptor, it is not doing any real analysis.

Have you experienced any issue? or do you have any more information on the matter?

shalugin:
> Keep in mind that the warning is shown because of the metadata descriptor, it is not doing any real analysis.

Yes, I know it,
> Have you experienced any issue? or do you have any more information on the matter?

I don't have any issue. I just wanted to know is the plugin thread safe.

Thank you for answer!

timkrueger:
It would be awesome if you can mark the asciidoctor-maven-plugin as thread safe. That will be remove a lot of unneeded logging in our build.

PR #363, opened 2018-08-04
Added new feature to fail based on different conditions

abelsromero:
Addresses: https://github.com/asciidoctor/asciidoctor-maven-plugin/issues/261, https://github.com/asciidoctor/asciidoctor-maven-plugin/issues/178

_Features are complete, but there's an issue with duplicated asciidoctor messages https://github.com/asciidoctor/asciidoctorj/issues/669._ Depending on suggestions we can merge or review the feature.

**Feature summary**
This PR adds a new configuration element as follows:
```xml
<logHandler>
	<outputToConsole>true</outputToConsole>
	<failIf>
		<severity>WARNING</severity>
		<containsText>super-important-document.adoc</containsText>
	</failIf>
</logHandler>
```
* `outputToConsole`: redirects asciidoctor messages to maven's logger. Messages are shown as Maven's INFO during renderitzation. Messages will appear twice as mentioned in issue https://github.com/asciidoctor/asciidoctorj/issues/669.
* `severity`: named so to align with AsciidoctorJ property. Will make the build fail if a message of this severity level is found. ~~It only matches the exact level, not same and above. I did this for simplicity but I am open to suggestions, implementation is trivial.~~ I finally decided to filter all messages above certain level to allow capturing all messages easily, also, feels more natural to understand.
* `containsText`: named so instead of `matchesText` because it just asserts that, a `contains`. I considered the option of allowing regex, but this made configuration more complex and less expressive for simple cases like a simple contains. Adding regex options could be done with another option or some custom syntax in the field.

Note that if both  `severity` and `containsText` are set, only messages that match both are captured.
Messages that match `severity` and/or `containsText` are shown as Maven ERROR (independently from `outputToConsole`), so that CIs can capture them as cause of build fail. Also, an initial error message showing the number of issues found is returned.

If someone wants to create a custom report like mentioned in https://github.com/asciidoctor/asciidoctor-maven-plugin/issues/261, you can register a custom LogHandler using the SPI interface (added integration test for that). However, this has some limitations with some formats that need to "close" and estructure (see https://github.com/asciidoctor/asciidoctor/issues/2832).

**Other notes**

* Filenames are formated to show relative path from project sourceDirectory. I considered that showing the absolute path as provided by AsciidoctorJ logger bloates the messages. Showing full path could be enabled with a new boolean option in the <logHandler> sections if requested.
* Since `severity` and `containsText`  are validated after the renderitzation the target document (html, pdf, etc.) can be found in the target directory. I don't see any reason to delete it.
[...]

PR #364, opened 2018-08-15
Mark BuildMojo as threadSafe

timkrueger:
This annotation will remove a maven build warning.

Resolves: #326

### DIFF
--- a/src/it/thread-safe/asciidoctor-project-1/pom.xml
+++ b/src/it/thread-safe/asciidoctor-project-1/pom.xml
@@ -35,6 +35,7 @@
               <sourceDirectory>src/main/doc</sourceDirectory>
               <outputDirectory>target/docs</outputDirectory>
               <backend>html</backend>
+              <headerFooter>false</headerFooter>
             </configuration>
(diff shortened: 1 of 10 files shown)


abelsromero:
We could try a `it` test with a couple of submodules with differnet documents.
If the documents don't mix content, we can assume it's OK (here there's an example https://github.com/abelsromero/asciidoctor-maven-plugin/tree/add_capture_of_Asciidoctor_log_messages/src/it/spi-registered-log).
Reviewing Maven documentation & code again, I don't see why it should fail un multi-thread scenario.

timkrueger:
I think now I got it 😁

abelsromero:
I guess that should be enough 👍

timkrueger:
@abelsromero Any plans when 1.5.7 will be released?

abelsromero:
Only pending thing is https://github.com/asciidoctor/asciidoctor-maven-plugin/pull/363. Since is a long requested feature I'd really appreciate some feedback :% but I'd probably merge this weekend regardless.

PR #367, opened 2018-09-25
closes #366 Control when cursor in logger is null

### LINKED ISSUE #366: `NullPointerException` when enabling logHandler.outputToConsole
_version: 1.5.7_

Some Asciidoctor messages cause a `NullPointerException` when `outputToConsole` is set to true.

This is due to the fact that this messages does not contain "cursor" element and the formatter fails.

See https://github.com/asciidoctor/asciidoctor-maven-plugin/pull/363#issuecomment-420963429

cc @lpetit-yseop

abelsromero:
Adds a null check to the cursor of logRecords.
In case the message does not have a cursor, it will print a message without it following Asciidoctor current format.

Example:
`[info] asciidoctor: WARN: invalid reference: section-id`


mojavelinux:
That's correct. Not all messages have a cursor because sometimes one isn't available.

PR #372, opened 2018-10-23
fix setting logHandler.outputToConsole to be true by default

abelsromero:


abelsromero:
It seems Travis is not supporting OracleJDK 10 now...but documentation states otherwise https://docs.travis-ci.com/user/languages/java/.

If this is the case, I'll merge and fix separatedly

PR #377, opened 2018-11-20
Update TravisCI setting to new vm based builds

abelsromero:
As mentioned here: https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

Travis is disabling docker model.
This PR just removes configurations that not necessary or working:

* removes `sudo: false`
* removes `dist: trusty`, which is not needed since trusty is already default

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,3 @@
-dist: trusty
-sudo: false
 language: java
 before_script:
   - export MAVEN_SKIP_RC=true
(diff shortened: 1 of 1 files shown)


mojavelinux:
I think we should still leave the `dist` line. While trusty may be the default today, when it becomes not the default, then we could run into a problem. Including it also increases transparency about what the build is actually using.

abelsromero:
restored

mojavelinux:
:+1: 

Btw, if Travis gives other advice about how to use the dist field, then of course we can follow it. Let's keep an eye out for that advice.

PR #384, opened 2018-12-30
Update asciidoctorj to 1.5.8.1

abelsromero:
This updates AsciidoctorJ version and tests that started failing due to https://github.com/asciidoctor/asciidoctor/issues/2722.

@mojavelinux I added about the issue in the README. I'd appreciate your opinion on wether we should make this visible or not.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -268,6 +268,12 @@ For example, set `include` to fail on any issue related to included files regard
 <2> Build will fail on any message of severity `DEBUG` or higher, that includes all.
 All matching messages will appear as ERROR in Maven output.
 
+[WARNING]
+====
+Due to link:https://github.com/asciidoctor/asciidoctor/issues/2722[issue #2722] in `asciidoctor`, versions >1.5.8.x of AsciidoctorJ cause internal reference warning messages not to be shown.
+Note however that using versions 1.5.7 and below may show false positives messages which may affect the build.
+====
+
 ==== Built-in attributes
 
 There are various attributes Asciidoctor recognizes.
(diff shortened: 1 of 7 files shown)


mojavelinux:
In my opinion, this should have never been enabled because it wasn't a complete feature. So we can mention it, but it's just consistent with the behavior of the Asciidoctor API now.

mojavelinux:
Btw, I think you can reenable this behavior by creating a script with the following contents:

```rb
$VERBOSE=true
```

Then requiring that script in the plugin configuration. That emulates the Asciidoctor API exactly.

abelsromero:
> Btw, I think you can reenable this behavior by creating a script with the following contents:

But then we may leak excessive information. I am more inclined to adding a <verbose> option and letting people choose.

mojavelinux:
> But then we may leak excessive information.

What I'm trying to say is that's the only way it worked before. This was never a warning that got passed through the logger under normal circumstances. So there's no loss of feature here. It works as it always has. (It just has to be enabled differently). Any change to have a proper warning is an enhancement for a future release.

abelsromero:
> It works as it always has. (It just has to be enabled differently).

Thanks for the clarification.
To avoid any impact on someone that may have been using this, even it was not intended, I will see if I can add `<enableVerbose>` option in another PR.

mojavelinux:
>  I will see if I can add <enableVerbose> option in this PR.

Sounds like a good idea.

abelsromero:
I must admit, it's been much, much easier than expected. The verbose script works fine even when setting and additional gempath, and when running from inside the plugin's jar.

I changed the warning to note and explained in a more positive way.

PR #389, opened 2019-03-29
Upgrade to AsciidoctorJ 2.0.0-RC.1

robertpanzer:
This PR makes the required changes to upgrade the plugin to AsciidoctorJ 2.0.0-RC.1

Notable changes are:
* Minimum required version of Java is now Java 8
* I ignored the tests for the warnings for invalid refs, I also have them commented in AsciidoctorJ

There's one test failing locally for me right now, which expects rendering with the Pygments highlighter to fail, however that succeeds.
I am unclear right now if this is now the expected behavior or if this is coming from my local machine setup, even though I don't have pygments installed locally afaict.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ jdk:
   - openjdk10
   - openjdk9
   - openjdk8
-  - openjdk7
 # skip installation step entirely
 install: true
 script:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,7 +4,6 @@ clone_depth: 10
 environment:
   MAVEN_VERSION: 3.5.0
   matrix:
-    - JAVA_HOME: C:\Program Files\Java\jdk1.7.0
     - JAVA_HOME: C:\Program Files\Java\jdk1.8.0
     - JAVA_HOME: C:\Program Files\Java\jdk9
 install:
--- a/pom.xml
+++ b/pom.xml
@@ -57,16 +57,16 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <project.java.version>1.7</project.java.version>
-        <project.execution.environment>JavaSE-1.7</project.execution.environment>
+        <project.java.version>1.8</project.java.version>
+        <project.execution.environment>JavaSE-1.8</project.execution.environment>
         <spock.version>0.7-groovy-2.0</spock.version>
         <groovy.version>2.1.3</groovy.version>
         <!--<asciidoctorj.version>1.6.0-SNAPSHOT</asciidoctorj.version>-->
-        <asciidoctorj.version>1.5.8</asciidoctorj.version>
+        <asciidoctorj.version>2.0.0-RC.1</asciidoctorj.version>
         <!--<asciidoctorj.version>1.5.7</asciidoctorj.version>-->
         <maven.coveralls.plugin.version>4.3.0</maven.coveralls.plugin.version>
         <maven.jacoco.plugin.version>0.8.2</maven.jacoco.plugin.version>
-        <jruby.version>9.1.17.0</jruby.version>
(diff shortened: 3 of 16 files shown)


abelsromero:
> There's one test failing locally for me right now, which expects rendering with the Pygments highlighter to fail, however that succeeds.
> I am unclear right now if this is now the expected behavior or if this is coming from my local machine setup, even though I don't have pygments installed locally afaict.

It was there as a reminder of an Asciidoctor bug, if it's fixed, we can update the test. I'll have a look at it.

One thing I had in my bucket list was the jruby dependency. Should we change to `org.jruby:jruby` instead of `jruby-complete`?

robertpanzer:
> One thing I had in my bucket list was the jruby dependency. Should we change to org.jruby:jruby instead of jruby-complete?

Good idea, I'll update the PR

abelsromero:
> Good idea, I'll update the PR

You have some time.
To be honest wanted to do a last release ensuring compatibility with 1.5.8.1 & 1.6.2 before working on 2.0.0. I am counting everything should work and it would not take much time tomorrow.
I'll write an issue with the 2.0.0 plan asap

robertpanzer:
That's also a great idea :D 
I just had some spare time so could do a quick PR for this.
I don't mind if it will take more time to review, that's all good.

abelsromero:
I already added the fix for the remote include test in master and confirmed the pygments issue is fixed in 2.x, but not before.

abelsromero:
I am merging, even when some test may be failing, I'll fix them right now.

abelsromero:
I checked the test that now passes with `pygments` and it seems that the previous error message `Pygments is not available` is not returned and the HTML is generated adding `pygments` css class to HTML code element. But not style is applied, not sure what I missing since I haver python2 installed and I could correctly convert the doc (with stylng) with the ruby cli after installing `pygments.rb` gem. Could it be we need to add pygments to AsciidoctorJ jar?
@robertpanzer do you think it makes sense to treat it in AsciidoctorJ.

Right now I am setting the test to Ignore until I have more time.

abelsromero:
>     * I ignored the tests for the warnings for invalid refs, I also have them commented in AsciidoctorJ

I hate sounding like a manager, but honestly, "I understand this is a known issue and is being worked on?"

robertpanzer:
I just remembered. 
It’s no longer ties to the -verbose flag. 
Instead it’s necessary to set the global variable `$VERBOSE` in an extension to get the warnings for invalid refs. 
I can add such an extension and enable the test again tomorrow.

robertpanzer:
For reference: https://github.com/asciidoctor/asciidoctorj/pull/689

abelsromero:
> I can add such an extension and enable the test again tomorrow.

Don't worry. In fact let it to me, so I can play with it and document it better in the readme.

abelsromero:
Surprise, it's already done https://github.com/asciidoctor/asciidoctor-maven-plugin/pull/384 I just forgot.
I'll check why it is not working

mojavelinux:
I know the changes in core caused a bit of confusion about when invalid references are logged. To clarify, as of 2.0, they get logged when the logger threshold is INFO or DEBUG. See https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor/substitutors.rb#L776

mojavelinux:
In other words, we no longer use the $VERBOSE global for this. That's now reserved for runtime warnings, such as an unused variable (which is the normal way of using it).

robertpanzer:
I think I see what's going on.
Asciidoctor now checks the log level before logging. 
I have to implement the explicit methods for info, warn etc.
I think previously Logger.add was called.

abelsromero:
If I understand correctly is a thing for AsciidoctorJ then. Please, refer the PR or issue to track it.

PR #431, opened 2020-05-31
removes imagesDir mojo configuration in favour of attributes

### LINKED ISSUE #296: Remove default value for imagesdir configuration.
This default value (`"images"`) caused some issues where documents were no longer able to specify an `:imagesdir:` attribute. For details, see the comments in #294, which has been resolved by at least making the default overridable.

What should happen for version `1.6.0`:
- [ ] Rename the setting from `imagesDir` to `imagesdir` to correspond to the attribute name.
- [ ] Remove the default value from the maven plugin configuration as defaults (if any) should be specified by asciidoctor core and not the maven plugin.

abelsromero:
Closes #296 

This PR:
* Removes the `imagesDir` configuration property
* Updates tests accordingly
* Adds small example to README to show how attributes are set


abelsromero:
Note that any issue with overriding imagesdir value disappears with this since the value must be set explicitly.

PR #433, opened 2020-06-01
closes #254 set default sourceDir to 'src/docs/asciidoc', fallbacks to 'src/asciidoc' and 'src/main/asciidoc' if not found

### LINKED ISSUE #254: Align sourceDirectory with the Gradle plugin
As proposed by @mojavelinux ([twitter](https://twitter.com/mojavelinux/status/755469879755419648)) and here ([asciidoctor-gradle-plugin/issues/108](https://github.com/asciidoctor/asciidoctor-gradle-plugin/issues/108)), the sourceDirectory should be `src/docs/asciidoc` as opposed to `src/main/asciidoc`.

abelsromero:
This PR aligns `sourceDirectory` to better practices while mantaining backward compatibility.
* Adds a new component `SourceDirectoryFinder` responsible for checking the different paths available and deciding the right one.
* The log messages in case sourceDirectory is not found have changed accordingly. I will add these to the 2.0.0 migration guide.

Starting with this PR I want to introduce some changes in how the plugin is mantained:
 * Create decoupled components (e.g. the new `SourceDirectoryFinder`) to extract logic from the Mojo. Imo, the Mojo has become too "spaghetti".
* Promote simple unit-testing of such components to reduce build time and tests complexity. In other words, avoid the current approach of testing all through the Mojo.
* Using common Java tools for testing. The goal is making the project more welcoming to people not knowing groovy/spock and in a lower degree make build faster. While Groovy+Spock is great, Java has improved to the point that I don't miss them in my daily works.


abelsromero:
@mojavelinux I'd wonder if you have any opinion about the changes in style I would like to introduce.

If not, I'd merge in some days and continue working on the 2.0.0 release.

mojavelinux:
I like this strategy.

PR #434, opened 2020-06-07
Support log handler configuration for site generation

### LINKED ISSUE #400: Support `logHandler` configuration for site generation
The ability to fail on warning or error messages from asciidoctor is not supported when generating a site, only when using the main `process-asciidoc` mojo.

We should definitely support this.

We should consider:
- [ ] Refactor current methods used to read XML configuration in [AsciidoctorDoxiaParser](https://github.com/asciidoctor/asciidoctor-maven-plugin/blob/master/src/main/java/org/asciidoctor/maven/site/AsciidoctorDoxiaParser.java) in a separate component to ease testing and future refactoring. Should hide implementation as much as possible.
- [ ] Implement support of same options `logHandler` options as the main `process-asciidoc` mojo.

/cc @ptahchiev feel free to give any feedback if you think this does not cover you use case.

abelsromero:
Closes #400 

This PR adds the possibility to add "logHandler" options to abort a site generation based on certain conditions (e.g. some warning messages are detected).
This PR:
* Adds `SiteConversionConfigurationParser` and `SiteLogHandlerDeserializer` to help reading site configurations.
* Removes support for `templateDir`, users should use `templateDirs`.
* Refactors log processing logic into `LogRecordsProcessors` to be reused between Mojos and site module. As a note, it has some "nasty" code due to the fact that Doxia and Mojos use different logging interfaces and return different exceptions. I am happy with this for now. In deeper refactors this can be reviewed.
* Updates README with logHandler explanation and some additional reviews.

It also includes:
* Update of plexus.component.metadata plugin to v1.7 to support usage of Lambas.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -1,6 +1,7 @@
 = Asciidoctor Maven Plugin
 // Metadata
 :release-version: 1.5.8
+:maven-site-plugin-version: 3.9.0
 // Settings
 :idprefix:
 :idseparator: -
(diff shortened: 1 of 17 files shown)


abelsromero:
Ready for review

PR #456, opened 2020-07-13
fix coveralls.io setup to run on GH actions

abelsromero:
This PR:
* Fixes configuration to push coverall.io resuls from GH Actions build pipeline. Thanks to @mojavelinux for setting up the repo.
This is restricted to only the linux, Java11 build for no special reason other than this is the more similar to production environments.
* Applies fix to `coveralls-maven-plugin` to run on Java11.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -33,6 +33,4 @@ jobs:
         run: mvn -B -V -Prun-its clean verify jacoco:report
       - name: Publish coverage reports
         if: matrix.os == 'ubuntu-latest' && matrix.java == '11' && github.base.ref == 'master'
-        env:
-          COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
         run: mvn coveralls:report -DrepoToken=${COVERALLS_REPO_TOKEN}
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -15,12 +15,9 @@ jobs:
       max-parallel: 2
       matrix:
         java:
-          - 8
           - 11
         os:
           - ubuntu-latest
-          - windows-latest
-          - macos-latest
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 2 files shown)


abelsromero:
I am struggling to get a clean solution for coveralls.io. While it "works", GH Actions prevents secrets from running in PR from forks.
That makes GH Actions not very usefull since coverage difference will not be appreciated in PRs.

I can setup the coverage upgade on merge as alternative if we want to remove travis and appveyor.

mojavelinux:
The coveralls docs seem to indicate that the repo_token is only needed for private repositories. Is this actually true? https://docs.coveralls.io/supported-ci-services

abelsromero:
Could be true, but the coveralls-maven-plugin requires it or else fails. We could send a PR to change it or improve integration with GH actions, but given the status of some issues and PR in the repo, it does not seem to be active. For example, compatibility with Java11 has not been fixed even though there's a PR and an issue. Not criticizing, just stating the fact that if no one is going to merge and relase the changes, there's not much we can do there.

I also saw this https://github.com/marketplace/actions/coveralls-github-action, but we depend on https://github.com/coverallsapp/github-action/issues/22 since it requires LCOV format.

Third way is building a custom integration here, maybe some groovy script. But I'd rather keep a single Travis build only for coverage than start building custom build solutions.

mojavelinux:
> just stating the fact that if no one is going to merge and relase the changes, there's not much we can do there.

Indeed, that's a constraint we have to work with until something changes.

> I'd rather keep a single Travis build only for coverage than start building custom build solutions.

Seems like a reasonable approach, at least in the near term.

abelsromero:
Did some more research and some test to double check.
The only thing (other than custom implementation) is restricting the execution to main branch, or run in cron. In both cases we don't have analisi on PR which to me is key.

I am going to mantain a single build in Travis java 11 just for it :man_shrugging:

mojavelinux:
Nice work. I know this was quite a journey, but I think you arrived at a nice solution given the constraints.

PR #461, opened 2020-07-17
removes unnecessary @Parameter configuration from ExtensionConfiguration, Synchronization and Resources

abelsromero:
As part of some clean up tasks, this PR:
* Removes unnecessary `org.apache.maven.plugins.annotations.Parameter` usages in ExtensionConfiguration, Synchronization and Resources, and in AsciidoctorMojo.
This is because properties in beans cannot be configured via terminal with _-D_. I'll do more search while the PR is open, but could not find any official doc so far and tests where unsuccessful. Note that List of String or File can be assigned passing a comma separated list like `mvn -Dasciidoctor.requires=onegem,anothergem`

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -20,6 +20,7 @@ Improvements::
 Bug Fixes::
 
   * Remove Maven components from plugin descriptor (#450)
+  * Remove unnecessary maven's @Parameter configuration from ExtensionConfiguration, Synchronization and Resources (#461)
 
 Documentation::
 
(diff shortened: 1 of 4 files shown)


abelsromero:
The most cohesive explanation I found is in https://maven.apache.org/guides/plugin/guide-java-plugin-development.html#parameters.
This always references "properties" or "variables" in the Mojo. Based on that and testing I assume is not possible to map anything that is not a simple type or a collection of simple types with -D properties.

Issue #472, opened 2020-07-21
Support `auto-refresh` with minimal features to be usefull


abelsromero:
I addition to the stated in the title itself, https://github.com/asciidoctor/asciidoctor-maven-plugin/issues/468 cannot be fully completed without fixing the following issues in the `auto-refresh` mojo:

- [x] Apply same selection rules for sources as in AsciidoctorMojo
- [x] Refresh individual files instead of triggering whole rebuild
- [ ]  Synchronize resources (including `<resources>`)

PS: It crossed my mind to make refresh and http mojos options in the main mojo, so these can be enable with a cli param. That way users don't need to change anything and reuse exact same configuration for local writting and build. But it seems too drastic to me and counter intuitive with other maven plugins like site, wich has separated run and site mojos build and preview. But if anyone sees that worth while, we can thing about it for a quick 3.0.0 release after 2.1.0

Issues found
~~Current implementation applies invalid filter to delete target files. It is using source extensions, no target one. But raises the issue of what happens if uses have for example html as resources. So we cannot just delete all.~~
Deletion of sources and resources is not currently supported see [comment below](https://github.com/asciidoctor/asciidoctor-maven-plugin/issues/472#issuecomment-665303234)

abelsromero:
As a note, support to delete converted documents once the source is delete will not be supported.
I run some tests and the problem is that we don't know the target extension, so we enter into the territory of guessing with may cause issues.
For example, we may delete file(s) with same base name, but then we can delete unwanted files, imagine a case where each source has an image with the same name. Maybe unlikely, but I fear providing this feature may cause more harm than good.
If user wants to clean he can always use the "refresh" command or stop and re-run.

PR #476, opened 2020-07-29
Refactor/asciidoctor mojo to allow dynamic resources processing logic

abelsromero:

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Refactor
- [ ] Build improvement
- [ ] Other (please describe)

Refactors to allow implementing "file by file" resource processing in refresh mojo.
Even when does not fix or add anything, I think it's worth keeping this in the git history.

Yes. Eventually we want to better refactor the code to extract processing logic from AsciidoctorMojo.
Right now the code is kept clean but with minimal changes. One example is the use of the mojo as configuration holder which is not idea, but a beginning.

No

- [ ] Yes
- [x] No
Related with https://github.com/asciidoctor/asciidoctor-maven-plugin/issues/472 but not closes.

Since this is internal refactor I don't think is work mentioning explicitly in CHANGELOG.
Refactors will be mentioned as a general highlight in the release notes.

### DIFF
--- a/src/main/java/org/asciidoctor/maven/AsciidoctorMojo.java
+++ b/src/main/java/org/asciidoctor/maven/AsciidoctorMojo.java
@@ -251,7 +251,7 @@ public class AsciidoctorMojo extends AbstractMojo {
 
         final Set<File> uniquePaths = new HashSet<>();
         for (final File source : sourceFiles) {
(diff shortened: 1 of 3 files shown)


abelsromero:
Passing 80% coverage :fireworks:

PR #478, opened 2020-07-30
Adds synchronization of resources for refresh mojo

### LINKED ISSUE #472: Support `auto-refresh` with minimal features to be usefull

abelsromero:

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Refactor
- [ ] Build improvement
- [ ] Other (please describe)

Adds efficient synchronization of resources for refresh mojo.

`*FileAlterationListenerAdaptor` hierarchy has some duplication but its a tradeoff I think is worth. The other alterntive is making some complex solutions with several lambdas or create abstract methods to customize log messages.

Closes #472

- [x] Yes
- [ ] No

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -17,6 +17,7 @@ Improvements::
 
   * Inject Maven properties as attributes in `process-asciidoc` mojo (#459)
   * Make `auto-refresh` (and `http` by inheritance) only convert modified and created sources (#474)
+  * Make `auto-refresh` only copy modified and created resources + taking into consideration <resources> options (#478)
 
 Bug Fixes::
 
--- a/src/main/java/org/asciidoctor/maven/AsciidoctorHttpMojo.java
+++ b/src/main/java/org/asciidoctor/maven/AsciidoctorHttpMojo.java
@@ -40,14 +40,13 @@ public class AsciidoctorHttpMojo extends AsciidoctorRefreshMojo {
     @Parameter(property = PREFIX + "reload-interval", defaultValue = "0")
     protected int autoReloadInterval;
 
+
     @Override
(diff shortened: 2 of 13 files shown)


abelsromero:
Side note, with the new test, build time has increased considerably

abelsromero:
PR ready!

abelsromero:
The problem with coverage reports does not seem to be a random thing. A workarround is available, but we should try to see the real reason.
https://github.com/jacoco/jacoco/issues/1076

abelsromero:
Fixed the issues with jacoco. It required some changes in the http-mojo.
Just applied minimal changes to make it work properly. We can improve it in another PR.

abelsromero:
+3 % coverage :partying_face:

PR #482, opened 2020-08-30
automates application of licence header to source files with license-maven-plugin

abelsromero:

- [ ] Bugfix
- [] Feature
- [x] Documentation
- [ ] Refactor
- [ ] Build improvement
- [ ] Other (please describe)

~Add & automate copyright to all source files.~
Original intention changed as per discusion below, now:
* Add copyright notice to readme.
* Remove unnecessary licence headers. All those that comply with license set at project level.

~First, there's al anternative plugin (https://www.mojohaus.org/license-maven-plugin) but I found it seems less mantained.
Secondly, the year could be automated using `<maven.build.timestamp.format>yyyy</maven.build.timestamp.format>`, but I am afraid of possible side effects during release.~
No

No

- [ ] Yes
- [x] No

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -35,6 +35,7 @@ Documentation::
   * Fix broken link to V2 migration guide (https://github.com/ge0ffrey[@ge0ffrey]) (#446)
   * Add GitHub's PR and issue templates (#465)
   * Add `auto-refresh` mojo documentation (#466)
+  * Add copyright notice to README and remove unnecessary license header from sources (#482)
 
 Build / Infrastructure::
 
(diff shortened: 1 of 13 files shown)


abelsromero:
@mojavelinux I used the header of the gradle-plugin as reference but based on other recent experiences, I wonder if the copyright should be `the Asciidoctor project` instead of current `the original author or authors`.

abelsromero:
After reviewing the main repos that include copyright notice (asciidoctor, asciidoctor.js, asciidoctor-intellij-plugin, gradle-plugin, asciidoctor-browser-extension and asciidoctor-reveal.js), I will apply the formula:

`Copyright © 2013-YYYY the individual contributors and the Asciidoctor Project`.

Honestly, if I have to put specific names I don't know where to begin with, so many great ppl.

mojavelinux:
Technically, you don't need license headers on every source file. You only need it at the root of the project. That applies to all files in the repository. The only reason you need it per file is if that file carries a different license or needs to be reattributed.

Generally the pattern I have used is that the original creator of the project and the current maintainer get mentioned in the attribution and then we say "individual contributors" which is a reference to the commits in the history.

PR #535, opened 2021-07-07
Replace usage of jruby DirectoryWalker with Java NIO API (#532)

### LINKED ISSUE #532: Remove usage of internal Asciidoctorj DirectoryWalker
**What is this issue about?**

- [ ] Bug report
- [ ] Feature request
- [ ] Question
- [x] Refactor/Improvement

**Description**
Currently, we use `org.asciidoctor.jruby.AsciiDocDirectoryWalker` and `org.asciidoctor.jruby.DirectoryWalker` to obtain AsciiDoc sources from the paths. But his classes are not part of the asciidoctorj-api but the jruby implementation, this breaks encapsulation and prevents the plugin from working with other AsciidoctorJ implementations using the service loader.
We should replace this by some agnostic piece with same functionality, for example [FileUtils.listFiles](http://commons.apache.org/proper/commons-io/apidocs/index.html?org/apache/commons/io/FileUtils.html).

Reference: https://github.com/asciidoctor/asciidoctor-maven-plugin/blob/2a79b16fda4ad30f78ff86799dfd9ecac7cc832f/src/main/java/org/asciidoctor/maven/AsciidoctorMojo.java#L439-L450

PD: there are other jruby specifics being used, but we can treat them in another issue.

stdll:

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Refactor
- [ ] Build improvement
- [ ] Other (please describe)

Closes #532 by replacing the class in question with the Java NIO API

There are lots of 3rd party libraries to do similar stuff, but I didn't want to rely on those in favor of Java's own NIO API.

I'm not sure about the error handling at `SourceDocumentFinder` (line 70). The `DirectoryWalker` didn't throw an exception and I didn't want to introduce one, so I handled it that way. I don't know if this is what you'd expect. If you have any feedback on this, I'm happy to change it.

- [X] Yes
- [ ] No

Fixes #532


src/main/java/org/asciidoctor/maven/process/SourceDocumentFinder.java
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * Finds all source documents inside the source directory with custom file extensions.
+     *
+     * @param sourceDirectory          source directory
+     * @param sourceDocumentExtensions custom file extensions
+     * @return the list of all matching source documents.
+     */
+    public List<File> find(Path sourceDirectory, List<String> sourceDocumentExtensions) {
+        String extensionPattern = sourceDocumentExtensions.stream()
+            .reduce(CUSTOM_FILE_EXTENSIONS_PATTERN_PREFIX, (result, s) -> result.concat("|").concat(s))
```
abelsromero, line 51: I suspect this adds `|` at the beginning of each file extension, so the pattern ends looking like `...(|pattern1|pattern2|pattern3)...` which means files with empty extension (`file.`) should be returned.

We could try fixing with `for + java.util.StringJoiner` or joining collector :thinking: 
```suggestion
String extensionPattern = sourceDocumentExtensions.stream()
                .collect(Collectors.joining("|", CUSTOM_FILE_EXTENSIONS_PATTERN_PREFIX, CUSTOM_FILE_EXTENSIONS_PATTERN_SUFFIX));
```

src/test/java/org/asciidoctor/maven/process/SourceDocumentFinderTest.java
```diff
@@ -1,21 +1,22 @@
...
+import java.nio.file.Paths;
 import java.util.Collections;
 import java.util.List;
 
+import org.junit.jupiter.api.Test;
+
 import static org.assertj.core.api.Assertions.assertThat;
 
-public class CustomExtensionDirectoryWalkerTest {
+public class SourceDocumentFinderTest {
```
abelsromero, line 12: Previous tests trusted the underlying implementation and did focus on covering files/folders combinations. But if we replace it with custom implementation we should add more tests, for example, testing with empty and multiple extensions in the second param, non existing paths, etc.

CHANGELOG.adoc
```diff
@@ -37,6 +37,7 @@ Maintenance::
   * Rewrite `AsciidoctorHttpMojoTest` to Java to make it more approachable (#516)
   * Rewrite `AsciidoctorZipMojoTest` to Java to make it more approachable (#518)
   * Rewrite `AsciidoctorDoxiaParserTest` to Java + remove Groovy & Spock configurations (#519)
+  * Replace usage of JRuby `DirectoryWalker` with Java NIO API (https://github.com/stdll[@stdll]) (#532)
```
abelsromero, line 40: My fault for how I worded the issue :hand: The thing is technically it's an internal "Asciidoctorj" class from the implementation that uses jRuby (under the jruby package), not "JRuby" the library itself. So if we want to support other Asciidoctor implementation in the future, we should not use anything under "asciidoctor.jruby" packages. 
Does this help?

Here is more context https://github.com/asciidoctor/asciidoctorj/issues/1038.

abelsromero:
Thanks a lot for the PR, keeping this things is important! The only important thing is the `sourceDocumentExtensions` leaky pattern. Feel free to create a dedicated set of folders and files for it if you need to.

stdll:
I've updated the PR. I've integrated your solution to joining, added a few tests for file extensions and clarified the changelog entry.

abelsromero:
Great! Thanks a lot!

PR #536, opened 2021-07-07
Remove apache commons direct dependencies

abelsromero:
Removed because security alerts with commons-io:2.5 and
that the actual versions are out of our control because
they are provided by Maven runtime.

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Refactor
- [ ] Build improvement
- [ ] Other (please describe)

Reduce maintenance efforts on dependencies that are actually out of our control.
Both `commons-io` and `commons-lang3` are provided by Maven runtime, so even if we override for a newer versions we will still use the one in Maven.
So, it's better to just use the version provided.

No that I am aware.

The initial motivation was removing the security warning that github shows, but I don't think it will fix it.

- [ ] Yes
- [x] No

And I don't thinks it's relevant for the CHANGELOG.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,7 +71,6 @@
         <plexus.utils.version>3.0.23</plexus.utils.version>
         <plexus.component.metadata.version>1.7</plexus.component.metadata.version>
         <netty.version>4.1.59.Final</netty.version>
-        <commons.io.version>2.5</commons.io.version>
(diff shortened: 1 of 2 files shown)


abelsromero:
Seems to have worked and we don't get the securty alert in the welcome page.

Issue #403, opened 2019-04-26
Trigger auto-reload on included files changing


mrm1st3r:
When starting the http server using `mvn asciidoctor:http asciidoctor:auto-refresh` it will only trigger a refresh on exactly the files being built.
But when building large documentation as single-page HTML, the main source files mostly contain includes to other files. When those included files are changed, no auto-refresh is triggered.

Example:
-  `sourceDocumentExtension` is set to "index.adoc"
- "howtos.index.adoc" includes "howto/writing-documentation.adoc"
- "howto/writing-documentation.adoc" is being modified

Expected behavior: rebuild is triggered and "howtos.index.html" will be updated
Actual behavior: No rebuild is triggered and I still have to type in refresh into the console.

abelsromero:
The way it works is by adding a listener to the configured "sources" to minimize scanning all possible files. And I honestly would like to avoid scanning all files, not to mention that some people may use special file extensions or locations.

To do it right, we depend on https://github.com/asciidoctor/asciidoctor/issues/60. So, the only option I see right now is adding a parameter so that users can add a regex of additional files to scan using the command line. Something like `mvn asciidoctor:http asciidoctor:auto-refresh -Dmonitor=**/__includes/*.adoc`

mrm1st3r:
OK, I think I understand. I think that command line argument could also be set as a configuration for the plugin inside the pom.xml? It could just default to the same value that determines which files are being built, but be overridden individually by users.

Having to remember the correct regex and supplying it each time to the command might be not the best option to encourage people to write or update documentation more often. :)

abelsromero:
> OK, I think I understand. I think that command line argument could also be set as a configuration for the plugin inside the pom.xml

Can be done. About the semantics of the option, to make it simpler I think it would be better that the files defines by the regex are added to the documents configured through the other options. About the name, any suggestion?  I am between "listenTo" & "refreshOn"

mrm1st3r:
I would go for "refreshOn" because it is then more obvious that it is used for "auto-refresh".

abelsromero:
Adding this to next milestone. I experimented with IncludeProcessor, but no use, specially if someone is using a custom processors also.

abelsromero:
Closed with https://github.com/asciidoctor/asciidoctor-maven-plugin/pull/542

PR #542, opened 2021-07-18
Allow setting included sources to run full refresh on change

abelsromero:

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Refactor
- [ ] Build improvement
- [ ] Other (please describe)

Fix conversion not happending when included sources where modified in `autp-refresh` and `http` mojo.
As discussed in #403 at the end we added a new option available for those mojos to allow users to specify other files to "listen to changes on". On a change or creation, a full rebuild will be done.

Current  approach allows users to specify a set of extra sources to monitor.
On change a full rebuild (without copying resources) will happen. A full rebuild is necessary since we don't know which sources are affected.

1. Creating an extension (include processor or preprocessor) that creates a tree of dependencies to know exactly which files to modify.
Discarded because custom filters may interfere with other extensions used by users.
2. Manually parsing all sources as texts files and build such index manually.
Discarded because configurations that maniputale paths (with basePath for example) and have files with same names may be impossible to properly resolve or we may have to also resort to "full builds".

- [x] Yes
- [ ] No

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -36,4 +36,4 @@ jobs:
           key: ${{ runner.os }}-mvn-cache-${{ hashFiles('**/pom.xml') }}
           restore-keys: ${{ runner.os }}-mvn-cache-
       - name: Build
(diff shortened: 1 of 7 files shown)


abelsromero:
I had to increase the timeout for refresh tests for ubuntu-java8. I assume they run slower on than environment for some reason.

Issue #545, opened 2021-07-23
Logic change in how paths are validated


liampuk:
Thank you for taking your time to talk with us!

**What is this issue about?**

- [x] Bug report
- [ ] Feature request
- [ ] Question

**Description**

I'm having some issues with AsciiDoctor not being able to find `generated-docs` during my Jenkins build. This seems to be as Jenkins sometimes truncates the workspace folder in the path to start with `_`. Previously this was not a problem as asciiDoctor only checks the file name for an `_` prefix, but after #535 every part of the path is checked for for `_` prefixes.

[old code](https://github.com/asciidoctor/asciidoctor-maven-plugin/blob/2a79b16fda4ad30f78ff86799dfd9ecac7cc832f/src/main/java/org/asciidoctor/maven/process/CustomExtensionDirectoryWalker.java#L27) vs [new code](https://github.com/asciidoctor/asciidoctor-maven-plugin/pull/535/files#diff-1ddfad9b3410bbb92a97bf3928c6a3d6704ced7f7ec9ee40c27df87ebc515741R63)

an example path from a failed build:
`/var/lib/jenkins/workspace/_service_name_develop/service-name-service/src/docs/asciidoc`

Was this intentional? If not can this be reverted to the previous logic?

**Environment information**

* asciidoctor-maven-plugin version: 2.2.0

abelsromero:
> an example path from a failed build:
> `/var/lib/jenkins/workspace/_service_name_develop/service-name-service/src/docs/asciidoc`
> 
> Was this intentional? If not can this be reverted to the previous logic?

Resources outside of the project path is something we should not check. In the meantime is perfectly safe to use v2.1.0 of the plugin with latest Asciidoctorj v2.5.1.

abelsromero:
Since it can affect Jenkins users I am going to do a quick 2.2.1 release.

mojavelinux:
Is it accurate to say that the Maven plugin excludes files that begin with underscore by default? If so, perhaps this should be mentioned in the docs. I looked for it, but I couldn't find anything that clearly stated what the default behavior is. (This is related to this issue I came across this week: https://github.com/asciidoctor/asciidoctor.org/issues/800).

abelsromero:
For the maven plugin we have a note in the `sourceDirectory` https://docs.asciidoctor.org/maven-tools/latest/plugin/goals/process-asciidoc/#configuration
[...]

PR #550, opened 2021-07-26
Treat dot files and folders same as internal

abelsromero:

- [x] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Refactor
- [ ] Build improvement
- [ ] Other (please describe)

Seeing the recent light of an issue with internal files treatment #545 we have been able to spot similar issues with treatment of hidden (prefix "." files and directories).
For that reason I think it's reasonable the next changes:
- FIX: In v2.1.0 sources inside a hidden directory where not processed, but they are in 2.2.x
- FEATURE CHANGE: While in v2.1.0 resources folders with "." where copied, it's not consistent with sources and internal resources ("_"). So this PR applies same treatment as with "_" and aligns it with "." sources which are also ignored.

We could have left behauviour as in 2.1.0 and copy "." folders, but that's an odd case not consistent with the rest.

Users having resources in hidden folders "." will need to rename them.
This could impact Windows users that do not follow Unix "hidden" files conventions, but I struggle to argue using this naming is a good practice we want to promote.

- [x] Yes
- [ ] No

### DIFF
--- a/CHANGELOG.adoc
+++ b/CHANGELOG.adoc
@@ -17,7 +17,8 @@ For a detailed view of what has changed, refer to the {uri-repo}/commits/main[co
 
 Bug Fixes::
 
-  * Fix excluding sources when enclosing parent path starts with _ (#544)
+  * Fix excluding sources when enclosing parent path starts with _ (#546)
+  * Treat dot files and folders same as internal #5505)
 
 == v2.2.0 (2021-07-18)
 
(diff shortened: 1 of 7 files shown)


uniqueck:
Would you like to setup a invoker test or is that enough to cover this behaviour with normal module tests?

abelsromero:
I think this is should be enough. In fact one of the things I'd like to try for next 3.0.0 is refactor the code into more decoupled pieces and do a more unit test approach like in this case.
Most of the tests are in fact integration tests that rely on full asciidoctorj conversions and that adds slowness and complexity.

...but this is of course open to discussion.

PR #595, opened 2022-08-21
Split plugin and site modules

### LINKED ISSUE #574: Split maven plugin and site module in 2 separete submodules
**Description**

In order to better evolve the tools we need to split each set of features on its own modules.
There's quite confusions about the subset of features that we explain much better since the antora docs migration but spliting the code seems the logical step.

**Other considerations**
* As submodules under same parent we will still release two at the same time even if there are no changes in one of them.
* As a nice extra, users using the plugin won't be downloading doxia libraries which they don't need.
* Requisit for https://github.com/asciidoctor/asciidoctor-maven-plugin/issues/571

abelsromero:

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Refactor
- [ ] Build improvement
- [ ] Other (please describe)

Splig the Maven Plugin and Doxia components in different modules following Maven practices.
Current Jar works as both because we are generating the appropiate metadata for both use cases, but it requires users to download dependencies they don't need.

n/a

Published artifacts change, from single jar `org.asciidoctor:asciidoctor-maven-plugin` we will have 3:
* `org.asciidoctor:asciidoctor-maven-plugin`
* `org.asciidoctor:asciidoctor-doxia-module`
* `org.asciidoctor:asciidoctor-maven-commons` (shared dependency by both).

- [x] Yes
- [ ] No

Closes #574 


abelsromero:
Right now the doxia module is calle `asciidoctor-doxia-module`, but it may change.

abelsromero:
IMPORTANT: For history reasons I am not changing the project repo yet. I have seen IntelliJ not liking that the parent module is "asciidoctor-maven-tools" and the folder is "asciidoctor-maven-plugin", and furthermore there's sub-module called "asciidoctor-maven-plugin" ... but can be fixed renaming the folder.

abelsromero:
Completed successfull deployment (just staging) on Maven Central with new structure, this will be the layout, the new artifacts all appear under `org.asciidoctor`. I personally don't want to make it more complicated with extra "maven" sub-levels and doing so also means changing the artifact coordinates.

image

PR #597, opened 2022-08-26
Reimplemented resource copy using 'plexus.util.DirectorScanner' instead of maven-filtering

### LINKED ISSUE #335: Consider `DirectoryScanner` as alternative to copy resources using `maven-filtering`
I stumbled upon `org.codehaus.plexus.util.DirectoryScanner` (plexus-utils) which offers a way to select files using include and exclude patterns.
This will require some extra boilerplate code but will provide this advantages:
* Removing the use of  `org.apache.maven.model.Resource`. Preventing confusion about the not-supported maven filtering options that now appear on some IDEs that offer maven's XML auto-completion.
* Improve build time since the logic will be simplified.
* Reduce the project dependencies since `maven-filtering` won't be necessary. That way users will have less to download.

abelsromero:
In exchange for some extra file management code, this reduces the amount of dependencies
used and injected into AsciidoctorMojo. It should also speed up build times since we
are doing much less work now to copy resources.

* Created model classes for Resources to ensure Mojo descriptors don't leak internals
* Remove 'org.apache.maven.shared:maven-filtering' dependency
* Bump plexus-utils to v3.3.1
* Remove unnecessary IT 'resource-filtering' since "under-the-hood" filtering is no longer possible

closes #335 

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Refactor
- [ ] Build improvement
- [ ] Other (please describe)

- Simplify code, reduce dependencies (libraries and runtime components injected in AsciidoctorMojo).
- Disambiguate "maven filtering" (aka. content replacement) features that where nos supported but available if users knew the internals.

We could have have just exposed a custom model pojo to avoid leaking filtering. But I think is worth the effort given the simplification of dependencies and code.

No.

- [x] Yes
- [ ] No


abelsromero:
First pass implementation based on passing current tests, but I want to do some refactor and create a component for resource management that can be unit tested.

abelsromero:
Run some tests with a project with 20k resources and the time reduction at the end is marginal.

Issue #166, opened 2015-08-17
Replicate default maven-site table style


abelsromero:
_From @sikron on August 13, 2015 17:31_

Hi,
i want to use asciidoc for both my maven-site and a generated PDF. I followed both your 'asciidoc-maven-site-example' and 'asciidoctor-pdf-example' with no alterations.

The generated PDF looks really great, but the maven-site does not look like the PDF at all, e.g.
- the source-code is not formatted as source-code, only verbatim and without background
- table borders are invisible, cells do not have any background

Is this by intention? Can i customize this by myself - preferrable without editing the css, but e.g. by attributes in the pom.xml?

I added pictures for better understanding. this is the poor-looking maven-site:
![maven-site](https://cloud.githubusercontent.com/assets/2979076/9256779/c95012fa-41f1-11e5-94bd-34bcf1c0bddf.png)

and this is the great-looking PDF:
![pdf](https://cloud.githubusercontent.com/assets/2979076/9256816/2a9347ee-41f2-11e5-9c95-4abd45bc4de3.png)

_Copied from original issue: asciidoctor/asciidoctor-maven-examples#31_

abelsromero:
_From @sikron on August 13, 2015 20:43_

ok, i am answering my own question: 
in order to have good source-highlighting, etc, the asciidoctor-maven-plugin in version 1.5.2.1 is needed (i had 1.5.2)

abelsromero:
Glad you could find a solution.
Feel free to close the issue when you want :)

abelsromero:
_From @sikron on August 14, 2015 14:25_

i am not 100% happy with the theming actually, but right now it suffices, i think.
thanks!

abelsromero:
> i am not 100% happy with the theming actually, but right now it suffices, i think.

Please, let us know if there's anything that can be improved.

abelsromero:
_From @sikron on August 16, 2015 17:36_

Hi again, 
thanks for offering your help! :-)
Actually, i think, there is a problem with tables. I have made a dummy-maven-project for comparing the output of the HTML for tables with the plain maven-site-plugin and apt-format vs. the maven-site plugin with asciidoctor-maven-plugin 1.5.2.1. Comparing the table, the output for the plain apt generates a table with borders and alternating line-background-colors as expected. The output of the asciidoc-plugin generates a table without borders and background-colors.
Is this by intention? How can i let the asciidoc-table have borders and colors?

[...]

PR #603, opened 2022-09-29
Make doxia compliant module

### LINKED ISSUE #571: Create maven-site/doxia compliant module
**What is this issue about?**

- [ ] Bug report
- [x] Feature request
- [ ] Question

**Description**

Current maven-site/doxia integration (`AsciidoctorDoxiaParser`) converts sources to Html (without header/footer) using Asciidoctor and embeds the result as-is. As a result the styles applied do not match with the maven site skins (simply the `class` value do not match) and other differences appear as seen in https://github.com/asciidoctor/asciidoctor-maven-plugin/issues/166.
To generate a "maven-site valid" Html a proper module should be implemented that uses Asciidoctor AST alongside [Doxia's Sink](https://github.com/apache/maven-doxia/blob/master/doxia-sink-api/src/main/java/org/apache/maven/doxia/sink/Sink.java) so that the the generation of Html is handled by Doxia.

Pros:
* Standard maven-site HTML
* Users get a good out-of-the-box experince (no need for post-customizations)
* Users can use maven skins (e.g. https://maven.apache.org/skins/maven-fluido-skin/)

Cons:
* We can only support structures available in the Sink interface. For example, AsciiDoc admonitions of table captions are not available natively, but not much research has been done yet.

This is scheduled for v3.0.0 not for breaking changes but because as part of 3.0.0 we will split the project in sub-modules, that way we can provide multiple doxia modules, one with current approach and another one with this.

abelsromero:

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Refactor
- [ ] Build improvement
- [ ] Other (please describe)

Provide a doxia module that nicelly integrates with maven site HTML.
Current doxia module injects HTML generated code "as is" which does not take into consideration maven site css. Users can still add their own css, but those looking for an easy integration or only want to deal with writting content cannot use certain constructs as seen by issues like https://github.com/asciidoctor/asciidoctor-maven-plugin/issues/166.

closes #571 

Yes. If the goal was to fix specific issues (like tables) we could use the current approach adding asciidoctor templates. However, I found more interesting to delve deeper, and offer a different module so users can choose either use all the goodies of Asciidoctor html that requires extra integration (custom css), VS have an easier integration sacrificing* some Asciidcotor features.
 * Not all features can be supported easily since there's no equivalent in the Doxia Sink and maven site skins don't offer styling, for example admonitions. I hope to be able to offer something eventually, but won't be as nice.

About the implementation:
This PR adds a new module `asciidoctor-parser-doxia-module` (and renames previous one to `asciidoctor-converter-doxia-module`).
The new one, instead of injecting html as generated by asciidoctor, uses the AST to generate a better adapted HTML either using the provided maven site Sink, or building custom HTML when necessary. In effect it's an Asciidoctor Converter but using a custom architecture to traverse the AST.

No.
The previous doxia-module is still available and will be maintained.
Changes will be covered in "v2 to v3 migration guide".

- [x] Yes
- [ ] No

~I am opening as draft, the implementation is complete for this release, pending docs to be added to this same PR.
Docs will include what Asciidoctor features are exactly covered by this module.~
Done :tada: There are still lots of features missing in the new parser site module, but I need to put a limit to the scope for the next release.


abelsromero:
A sample with default skin 
image

abelsromero:
Sample with fluido skin
image

PR #629, opened 2023-03-19
Set minimal maven version to 3.9.x & bump dependencies

abelsromero:
Bump dependencies:
* maven-core v3.9.1
* commons-io v2.11.0
* plexus-utils v3.5.1

Plugins
* maven-plugin-plugin v3.8.1
* maven-plugin-annotations v3.8.1
* plexus-component-metadata v2.1.1
* maven-enforcer-plugin v3.2.1
* maven-compiler-plugin v3.11.0
* maven-release-plugin v3.0.0-M7 (validated it works fine with 2.2.3 release)
* maven-javadoc-plugin v3.5.0
* maven-surefire-plugin v3.7.0
* maven-invoker-plugin v3.4.0 (there's a known issue preventing v3.5.0 use)

### DIFF
--- a/.mvn/wrapper/maven-wrapper.properties
+++ b/.mvn/wrapper/maven-wrapper.properties
@@ -14,5 +14,5 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.8.6/apache-maven-3.8.6-bin.zip
+distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.9.1/apache-maven-3.9.1-bin.zip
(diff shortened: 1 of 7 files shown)


abelsromero:
Back to "draft" status while looking at the issues with integration tests.

abelsromero:
Found that `org.codehaus.plexus:plexus-utils` is not being added to ITs classpath even when it's part of the provided libraries of Maven. Setting as `compile` scope works but it should not be necessary.

abelsromero:
The issue seems a side-effect

abelsromero:
The issue seems a side effect of this [change](https://maven.apache.org/docs/3.9.0/release-notes.html) 
image


See https://issues.apache.org/jira/browse/MNG-6965.

For new we are going to set `plexus-utils` as `compile` dependency and keep track of the fix (or lack of it) in upstream Maven.

abelsromero:
It's been confirmed in https://issues.apache.org/jira/browse/MNG-6965 that this is not an issue, and `plexus-utils` must be set as `compile` :+1:

PR #669, opened 2023-12-26
Add Dependabot for non pinned dependencies

abelsromero:

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Refactor
- [x] Build improvement
- [ ] Other (please describe)

Automate maintenance of dependencies.
Dependabot is not perfect, but having used it for some time I find it helps. Even if the bump breaks something, at least we get a notification and we see that in CI and can decide what to do. The alternative is depending on time or strength to go through bumps and hope for the best.

There are others. I only have first-hand knowledge [Renovate](https://github.com/renovatebot/renovate), which I don't like for some companiy and data collection policies. In general, alternatives depend on extra configurations and authorizations, Dependabot being integrated in GitHub makes things easier.
Note: dependabot does text parsing, it does not really run a dependency analysis, which causes some errors, but it helps most of the times.

Pros compensate cons, but there are some:
* Changelog is not updated, we'd need to document the AsciidoctorJ & JRuby versions on the release notes.
* Maven and Java versions in CI are not covered.

Not for users.

- [ ] Yes
- [x] No

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,40 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "dependencies"
+  - package-ecosystem: "maven"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "dependencies"
+    commit-message:
+      prefix: "(main)"
+    target-branch: "main"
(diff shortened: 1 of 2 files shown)


abelsromero:
I ran a test on a forked repo and immediately got image.

It does not cover all dependencies (missing some plugins), but shows that plugins and dependencies are handled. I expect the others to get PRs eventually in my experience.

PR #725, opened 2024-01-01
Test artifacts signature

abelsromero:

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Refactor
- [x] Build improvement
- [ ] Other (please describe)

Verify that the generation of release artifacts (jars + signatures) works correctly.
Especially with Dependabot bumping plugins, we could break the deployment process without knowing it because we don't test the signature.

We could do a full deployment (except pushing to Maven central), but I want to go step, by step. At least, for now, this removes a risk for releases.

Note that I created a `test` environment with a test key. Just in case, this will be different from the one used for actual releases.

no

- [ ] Yes
- [x] No

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -84,6 +84,11 @@ jobs:
           - 3.9.6
     runs-on: ${{ matrix.os }}
     steps:
+      - name: debug
+        run: |
+          echo $GPG_KEYNAME
+          echo $GPG_PASSPHRASE
+          echo $GPG_PRIVATE_KEY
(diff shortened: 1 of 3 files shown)


abelsromero:
I see we can't use secrets on PRs coming from outside the repo for security reasons https://securitylab.github.com/research/github-actions-preventing-pwn-requests/. It make sense, we can explore using environment variables for this case since the key is something randomly generated and not published.

abelsromero:
Closing while working on alternatives in fork repo

PR #731, opened 2024-01-05
(v2.2.x) Return maven-filtering to v3.3.0 from v3.3.1

abelsromero:

- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Refactor
- [ ] Build improvement
- [x] Other (please describe)

Avoid upstream bug.

v3.3.1 introduced a change that causes confusing output when using 'project.baseDir' for resources path.
See https://issues.apache.org/jira/browse/MSHARED-1175

The old message when copying resouces is fine.
```
Copying 118 resources
```

But the new, while less informative, omits the source after "from" in some cases.
```
Copying 118 resources from  to target/generated-docs
```

No. I don't think we are loosing anything with the revert.

- [ ] Yes
- [x] No

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,7 +80,7 @@
         <maven.jacoco.plugin.version>0.8.11</maven.jacoco.plugin.version>
         <maven.plugin.plugin.version>3.9.0</maven.plugin.plugin.version>
         <maven.project.version>3.8.5</maven.project.version>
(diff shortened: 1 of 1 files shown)


abelsromero:
No documenting in changelog for release notes since this is a change to revert back to original behavior.

PR #841, opened 2024-05-15
(main) Fix resource copy in multi-module project (#838)

### LINKED ISSUE #838: Resource copy in multi-module does not work
**What is this issue about?**

- [x] Bug report 🪲
- [ ] Feature request 🏎
- [ ] Question ❓ Consider dropping it first in the [Project Chat](https://asciidoctor.zulipchat.com/#narrow/stream/users)

**Description**

Considering the following multi-module maven project:
* pom.xml
* module-doc\
    * pom.xml
    * src/docs/asciidoc/
        * index.adoc
        * images/sample.png

The `module-doc\pom.xml` contains the plugin execution with the following `resources`:
```
<resources>
	<resource>
		<directory>src/docs/asciidoc/images/</directory>
		<targetPath>images/</targetPath>
		<includes>
			<include>*.png</include>
		</includes>
	</resource>
</resources>
```

The `sample.png` image is not copied in the output directory as expected.

**Environment information**

* asciidoctor-maven-plugin version: 3.0.0
* asciidoctorj version: 2.5.11
* Apache Maven 3.9.3
* Java version: 17.0.8.1
* Windows 11

schastlivcev:

- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Refactor
- [ ] Build improvement
- [ ] Other (please describe)

Fix resource copy in multi-module project

*None*

*None*

- [x] Yes, Fixes #838
- [ ] No

### DIFF
--- a/asciidoctor-maven-plugin/src/main/java/org/asciidoctor/maven/process/CopyResourcesProcessor.java
+++ b/asciidoctor-maven-plugin/src/main/java/org/asciidoctor/maven/process/CopyResourcesProcessor.java
@@ -60,13 +60,13 @@ public class CopyResourcesProcessor implements ResourcesProcessor {
     @Override
     public void process(File sourceRootDirectory, File outputRootDirectory, AsciidoctorMojo configuration) {
         final List<Resource> finalResources = prepareResources(sourceRootDirectory, configuration);
-        copyResources(finalResources, outputRootDirectory);
+        copyResources(finalResources, outputRootDirectory, configuration);
     }
 
     /**
      * Initializes resources attribute excluding AsciiDoc documents, internal directories/files (those prefixed with
      * underscore), and docinfo files.
-     * By default everything in the sources directories is copied.
+     * By default, everything in the sources directories is copied.
      *
      * @return Collection of resources with properly configured includes and excludes conditions.
      */
@@ -116,14 +116,15 @@ public class CopyResourcesProcessor implements ResourcesProcessor {
      *
      * @param resources       Collection of {@link Resource} defining what resources to {@code outputDirectory}.
      * @param outputDirectory Directory where to copy resources.
+     * @param configuration   Project configuration
      */
-    private void copyResources(List<Resource> resources, File outputDirectory) {
+    private void copyResources(List<Resource> resources, File outputDirectory, AsciidoctorMojo configuration) {
 
         resources.stream()
-            .filter(resource -> new File(resource.getDirectory()).exists())
+            .filter(resource -> configuration.getBaseDir().exists())
             .forEach(resource -> {
                 DirectoryScanner directoryScanner = new DirectoryScanner();
-                directoryScanner.setBasedir(resource.getDirectory());
+                directoryScanner.setBasedir(configuration.getBaseDir());
 
                 if (resource.getIncludes().isEmpty())
                     directoryScanner.setIncludes(new String[]{"**/*.*", "**/*"});
--- a/asciidoctor-maven-plugin/src/main/java/org/asciidoctor/maven/process/CopyResourcesProcessor.java
+++ b/asciidoctor-maven-plugin/src/main/java/org/asciidoctor/maven/process/CopyResourcesProcessor.java
@@ -121,10 +121,21 @@ public class CopyResourcesProcessor implements ResourcesProcessor {
     private void copyResources(List<Resource> resources, File outputDirectory, AsciidoctorMojo configuration) {
 
         resources.stream()
-            .filter(resource -> configuration.getBaseDir().exists())
+            .map(resource -> {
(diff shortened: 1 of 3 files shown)


abelsromero:
ThanKs! I'll have a look as soon as possible, it may take a couple of days though 🙇

asciidoctor-maven-plugin/src/main/java/org/asciidoctor/maven/process/CopyResourcesProcessor.java
```diff
@@ -116,14 +116,15 @@ private List<Resource> prepareResources(File sourceDirectory, AsciidoctorMojo co
...
      * @param resources       Collection of {@link Resource} defining what resources to {@code outputDirectory}.
      * @param outputDirectory Directory where to copy resources.
+     * @param configuration   Project configuration
      */
-    private void copyResources(List<Resource> resources, File outputDirectory) {
+    private void copyResources(List<Resource> resources, File outputDirectory, AsciidoctorMojo configuration) {
 
         resources.stream()
-            .filter(resource -> new File(resource.getDirectory()).exists())
+            .filter(resource -> configuration.getBaseDir().exists())
```
abelsromero, line 124: I think that what we want here is `getProjectDirectory()`, thinking about it is not as simple as I initially though.

There's the "project's baseDir", that's the path in which the pom resides (this is set always by maven using this [annotation here](https://github.com/asciidoctor/asciidoctor-maven-plugin/blob/5077f0425b9cae3dff6f37039fcdb6e1a5c6caf9/asciidoctor-maven-plugin/src/main/java/org/asciidoctor/maven/AsciidoctorMojo.java#L73)). And also "Asciidoctor's baseDir" which is an option to set the root where Asciidoctor searches for files (likes includes).
The latter (which is being used in this code) can be null, and I don't think we should deal with it. It's my assumption that in `<resources>` the users will want to put either an absolute path or a relative path from the pom's location.

That means we should check if the path in the `<resource>` is absolute or not.
If it's absolute, we do the same we do now, if it's not, we need to compose that with `configuration.getProjectDirectory()`. On way to do it would be adding creating a new Resource instance before the filter using `map()`.
```java
            .map(resource -> {
                final File candidate = new File(resource.getDirectory());
                final File sourceDir = candidate.isAbsolute() ? candidate : new File(configuration.getProjectDirectory(), resource.getDirectory());

                final Resource sanitizedResource = new Resource();
                sanitizedResource.setDirectory(sourceDir.getAbsolutePath());
                sanitizedResource.setTargetPath(resource.getTargetPath());
                sanitizedResource.setIncludes(resource.getIncludes());
                sanitizedResource.setExcludes(resource.getExcludes());
                return sanitizedResource;
            })
```

I haven't tested myself, but I think it should work :thinking: so all these work
```
    <resource>
        <directory>src/docs/asciidoc/images/</directory>
        <targetPath>images/</targetPath>
        <includes>
            <include>*.png</include>
        </includes>
    </resource>
```

```
    <resource>
        <directory>${project.basedir}/src/docs/asciidoc/images/</directory>
        <targetPath>images/</targetPath>
        <includes>
            <include>*.png</include>
        </includes>
    </resource>
```


Btw, if a user inputs `<directory>/src/docs/asciidoc/images/</directory>`, to me that's a configuration error and it should be considered an absolute path. We shouldn't try to guess and interpret `/`.

abelsromero (changes requested):
Thanks a lot for the time, but I am really sorry to say I don't think the approach would not work.
The IT tests are not easy, I can take care of that later if you want. But you should be able to try it in your project installing the plugin locally. Run `mvn clean install -DskipTests` and configure your project to use version `3.0.1-SNAPSHOT`.

asciidoctor-maven-plugin/src/main/java/org/asciidoctor/maven/process/CopyResourcesProcessor.java
schastlivcev: Oh... I see. I'll try)

schastlivcev: Tried, seems working

abelsromero: Thanks, I am quite busy, so I'll check first thing dueing the weekend.

abelsromero:
Thanks, all is fine, I double-checked again branch v2.2.x and we don't need to port it. So we are set.
Since there's a simple workaround I am planning a release asap, but feel free to object if you need this.
I'll start work on the labeled [bugs](https://github.com/asciidoctor/asciidoctor-maven-plugin/issues?q=is%3Aissue+is%3Aopen+label%3Abug), but those are available for anyone to contribute.

Issue #1023, opened 2025-02-16
Allow logging to a file (besides logging to the console)


dhendriks:
According to the documentation at https://docs.asciidoctor.org/maven-tools/latest/plugin/goals/process-asciidoc/#configuration-logHandler, we can configure the `logHandler`. We can configure `outputToConsole`, by setting it to `true` or `false`. However, I'd like to log to both the console and to a file (to generate a composite report of all problems of various Maven plugins/goals). Could an option named `outputFile` be added for the `logHandler`, that gives the path to the output file, such that logging is (also) written there?

What do you think?

abelsromero:
I assumed this could be done with common shell tools like `tee`. Isn't that an option?

abelsromero:
Not saying no outright, but outputting to a file won't work. I want to make sure the feature makes sense in the long term.

The main problem is that the current feature is not thought of as a reporting tool, instead, to fail-fast builds. We collect the messages for the latest converted doc and fail if there are any. If you have 10 files, and the first one is the one with errors, that's the only information you are going to get.
And I understand that's not the desired behavior here. Something structured should be generated, no need for HTML. But at least XML or JSON? (like junits).

The motivation was to ensure that if the build succeeds, it contains no errors (based on the configuration). In that regard, you can run `mvn verify` and be confident all is OK.

My assumptions for that approach were:
1. Most errors should have been caught during writing.
IDE plugins provide help for that. Or you can use `auto-refresh`, `http` plugin goals to get faster feedback loops with a 100% accurate conversion. Here, the user knows what is being modified and there's no need to report or convert all docs, any error should be related to something recently modified.
3. `process-asciidoc` should be run for verification and final conversion in automated environments.
I see how the current fail-fast is an issue, and having an option to not fail-fast makes sense, to be able to see all issues and fix them at once. But that's still a process that's based on plain output, not restructured reports.

I think those still hold. Can you explain more about the writing workflow? I don't mean to be rude, but I don't think the user is writing 3 or 4 pages, running a conversion, and then looking at a xml/json report 🙇 

PD: there's some mention of multi-modules, I am not considering that, unless we go the rabbit hole of creating a reporting via maven-site. But that would complicate configuration for users imo.

dhendriks:
Thanks for listing your thoughts on this! Here are some replies to the points you raise:

[...]

PR #1032, opened 2025-03-08
Add failFast option to allow displaying all errors in build

abelsromero:

- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Refactor
- [ ] Build improvement
- [ ] Other (please describe)

Allows to see all errors in all files before reporting a build failure. Previously, build failed on the first error found.

See https://github.com/asciidoctor/asciidoctor-maven-plugin/issues/1023

The approach opts to extend the current features for minimal changes.
See the whole conversation in the linked issue for context.

No.
The default behavior remains the same.

- [x] Yes
- [ ] No

### DIFF
--- a/asciidoctor-maven-commons/src/main/java/org/asciidoctor/maven/log/CapturedLogRecord.java
+++ b/asciidoctor-maven-commons/src/main/java/org/asciidoctor/maven/log/CapturedLogRecord.java
@@ -1,0 +1,66 @@
+package org.asciidoctor.maven.log;
+
+import java.io.File;
+import java.util.Optional;
+
+import org.asciidoctor.ast.Cursor;
+import org.asciidoctor.log.LogRecord;
+
+/**
+ * {@link LogRecord} proxy that allows capturing the source file being
+ * processed.
+ * Important: the {@link #sourceFile} and the actual source where an error is present
+ * may not be the same. For example if the source is being included.
+ *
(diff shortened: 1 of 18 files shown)


asciidoctor-maven-commons/src/main/java/org/asciidoctor/maven/log/CapturedLogRecord.java
```diff
@@ -0,0 +1,66 @@
...
+    private final File sourceFile;
+
+    CapturedLogRecord(LogRecord record, File sourceFile) {
+        super(record.getSeverity(), record.getCursor(), record.getMessage(), record.getSourceFileName(), record.getSourceMethodName());
+        this.sourceFile = sourceFile;
+    }
+
+    public Cursor getCursor() {
+        if (sourceFile == null)
+            return null;
```
dhendriks, line 28: If `sourceFile` is `null`, but `super.getCursor()` returns something non-`null`, shouldn't this return the super cursor instead of `null`? It seems this overrides the super cursors, but should provide only a fallback?

abelsromero: 😮 true! thanks. That's what happens when I don't do  proper unit tests.

